PR #15, opened 2020-07-22
General improvements in case of misconfiguration

cyattilakiss:
## Summary
Don't break the checkout flow in case the origin key or the payment methods list is missing

## Tested scenarios

Wrong API key therefore origin key cannot be retrieved and paymentMethods cannot be retrieved

### DIFF
--- a/src/Storefront/Page/Checkout/Confirm/AdyenCheckoutConfirmPageLoader.php
+++ b/src/Storefront/Page/Checkout/Confirm/AdyenCheckoutConfirmPageLoader.php
@@ -128,12 +128,19 @@ class AdyenCheckoutConfirmPageLoader extends CheckoutConfirmPageLoader
             //If this is an Adyen PM installed it will only be enabled if it's present in the /paymentMethods response
             if (strpos($paymentMethodEntity->getFormattedHandlerIdentifier(), 'adyen')) {
                 $pmCode = 'scheme'; //TODO get from payment method handler instead of hardcoding PM type
+                // In case the paymentMethods response has no payment methods, remove it from the list
+                if (empty($adyenPaymentMethods)) {
+                    $originalPaymentMethods->remove($paymentMethodEntity->getId());
+                    continue;
+                }
+
                 $pmFound = array_filter(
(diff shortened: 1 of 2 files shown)


src/Storefront/Page/Checkout/Confirm/AdyenCheckoutConfirmPageLoader.php
```diff
@@ -128,12 +128,19 @@ private function filterShopwarePaymentMethods($originalPaymentMethods, SalesChan
             //If this is an Adyen PM installed it will only be enabled if it's present in the /paymentMethods response
             if (strpos($paymentMethodEntity->getFormattedHandlerIdentifier(), 'adyen')) {
                 $pmCode = 'scheme'; //TODO get from payment method handler instead of hardcoding PM type
+                // In case the paymentMethods response has no payment methods, remove it from the list
+                if (empty($adyenPaymentMethods)) {
+                    $originalPaymentMethods->remove($paymentMethodEntity->getId());
+                    continue;
+                }
```
acampos1916, line 135: Wouldn't this remove all payment methods from the list including Shopware defaults and other PMs installed?

src/Storefront/Page/Checkout/Confirm/AdyenCheckoutConfirmPageLoader.php
```diff
@@ -128,12 +128,19 @@ private function filterShopwarePaymentMethods($originalPaymentMethods, SalesChan
             //If this is an Adyen PM installed it will only be enabled if it's present in the /paymentMethods response
             if (strpos($paymentMethodEntity->getFormattedHandlerIdentifier(), 'adyen')) {
```
msilvagarcia, line 129:
```suggestion
            if (strpos($paymentMethodEntity->getFormattedHandlerIdentifier(), 'adyen') !== false) {
```

cyattilakiss: Makes sense, thanks for noticing!

PR #18, opened 2020-07-24
[PW-2716] Update or create statedata for token

acampos1916:
## Summary
This modification updates the state data record related to the context token in order to keep only one record.

## Tested scenarios
Generating a new statedata updates the PaymentStateData table for that token.
Generating a new statedata creates a new record if the token is not present.

**Fixed issue**:  PW-2716

### DIFF
--- a/src/Service/PaymentStateDataService.php
+++ b/src/Service/PaymentStateDataService.php
@@ -83,17 +83,14 @@ class PaymentStateDataService
 
         if ($stateData) {
             $fields['id'] = $stateData->getId();
-            $this->paymentStateDataRepository->update(
-                [$fields],
-                \Shopware\Core\Framework\Context::createDefaultContext()
-            );
-        } else {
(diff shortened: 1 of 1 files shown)


src/Service/PaymentStateDataService.php
```diff
@@ -79,18 +79,27 @@ public function insertPaymentStateData(string $contextToken, string $stateData,
...
+            $this->paymentStateDataRepository->update(
+                [$fields],
+                \Shopware\Core\Framework\Context::createDefaultContext()
+            );
+        } else {
+            $this->paymentStateDataRepository->create(
+                [$fields],
+                \Shopware\Core\Framework\Context::createDefaultContext()
+            );
+        }
```
msilvagarcia, line 95: Can we use `EntityRepository#upsert` instead?

acampos1916: done!

PR #19, opened 2020-07-28
[PW-2739] Handle Error and Rejected result codes

cyattilakiss:
## Summary
Add PaymentResponseHandler to CardsPaymentMethodHandler

## Tested scenarios
Rejected card payment

### DIFF
--- a/src/Handlers/CardsPaymentMethodHandler.php
+++ b/src/Handlers/CardsPaymentMethodHandler.php
@@ -31,12 +31,11 @@ use Adyen\Service\Builder\Customer;
 use Adyen\Service\Builder\Payment;
 use Adyen\Service\Validator\CheckoutStateDataValidator;
 use Adyen\Shopware\Service\PaymentStateDataService;
-use Shopware\Core\Checkout\Payment\Cart\AsyncPaymentTransactionStruct;
-use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\AsynchronousPaymentHandlerInterface;
-use Shopware\Core\Checkout\Payment\Exception\AsyncPaymentProcessException;
+use Shopware\Core\Checkout\Payment\Cart\SyncPaymentTransactionStruct;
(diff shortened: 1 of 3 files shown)


src/Handlers/CardsPaymentMethodHandler.php
```diff
@@ -401,7 +414,7 @@ public function preparePaymentsRequest(
             ),
             $transaction->getOrder()->getOrderNumber(),
             $this->configurationService->getMerchantAccount(),
-            $transaction->getReturnUrl(),
+            'test url',
```
msilvagarcia, line 417: Expected?

src/Handlers/PaymentResponseHandler.php
```diff
@@ -26,14 +26,18 @@
...
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Shopware\Core\Checkout\Payment\Cart\AsyncPaymentTransactionStruct;
+use Shopware\Core\Checkout\Payment\Exception\AsyncPaymentProcessException;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 
 class PaymentResponseHandler
 {
-    const ADYEN_MERCHANT_REFERENCE = 'adyenMerchantReference';
+    const ADYEN_MERCHANT_REFERENCE = 'merchantReference';
```
acampos1916, line 40: This must match the GET var the is returned in the URL, in PS is adyenMerchantReference. Did you change it for some different reason?

cyattilakiss: aaaa sorry so this is for the GET parameter - I use it for the synch response merchantReference var
I will create a new one!

src/Handlers/CardsPaymentMethodHandler.php
cyattilakiss: Thanks for noticing, I will put back the original one and if it's necessary we can adjust it later to a new url

PR #20, opened 2020-07-28
[PW-2741] Ajax payments call from checkout

acampos1916:
## Summary
Payment API endpoint.
Confirm order JS plugin to make Ajax payment call to the backend.
Renamed migration class for consistency.
Renamed event subscriber class to be generic.

## Tested scenarios
NA

**Fixed issue**:  [PW-2741]

### DIFF
--- a/src/Controller/SalesChannelApiController.php
+++ b/src/Controller/SalesChannelApiController.php
@@ -140,7 +140,7 @@ class SalesChannelApiController extends AbstractController
      * @param SalesChannelContext $context
      * @return JsonResponse
      */
-    public function payment(SalesChannelContext $context): JsonResponse
+    public function getPaymentStatus(SalesChannelContext $context): JsonResponse
     {
         return new JsonResponse(['aloha']);
     }
--- a/src/Controller/SalesChannelApiController.php
+++ b/src/Controller/SalesChannelApiController.php
@@ -128,4 +128,20 @@ class SalesChannelApiController extends AbstractController
     {
         return new JsonResponse($this->paymentDetailsService->doPaymentDetails($context));
     }
+
+    /**
+     * @RouteScope(scopes={"sales-channel-api"})
+     * @Route(
+     *     "/sales-channel-api/v1/adyen/payment",
+     *     name="sales-channel-api.action.adyen.payment",
+     *     methods={"POST"}
+     * )
+     *
+     * @param SalesChannelContext $context
+     * @return JsonResponse
+     */
+    public function payment(SalesChannelContext $context): JsonResponse
+    {
+        return new JsonResponse(['aloha']);
+    }
 }
--- a/src/Controller/SalesChannelApiController.php
+++ b/src/Controller/SalesChannelApiController.php
@@ -132,9 +132,9 @@ class SalesChannelApiController extends AbstractController
     /**
      * @RouteScope(scopes={"sales-channel-api"})
      * @Route(
-     *     "/sales-channel-api/v1/adyen/payment",
-     *     name="sales-channel-api.action.adyen.payment",
-     *     methods={"POST"}
+     *     "/sales-channel-api/v1/adyen/payment-status",
+     *     name="sales-channel-api.action.adyen.payment-status",
+     *     methods={"GET"}
      * )
      *
      * @param SalesChannelContext $context
--- a/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
@@ -1,0 +1,55 @@
+import Plugin from 'src/plugin-system/plugin.class';
+import DomAccess from 'src/helper/dom-access.helper';
+import StoreApiClient from 'src/service/store-api-client.service';
+import FormSerializeUtil from 'src/utility/form/form-serialize.util';
+import ElementLoadingIndicatorUtil from 'src/utility/loading-indicator/element-loading-indicator.util';
+
+/* global adyenCheckoutOptions */
+export default class ConfirmOrderPlugin extends Plugin {
+
+    init() {
+        const confirmOrderForm = DomAccess.querySelector(document, '#confirmOrderForm');
+        confirmOrderForm.addEventListener('submit', this.confirmOrder.bind(this));
+    }
+
+    confirmOrder(event) {
+        event.preventDefault();
+        const form = event.target;
+        if (!form.checkValidity()) {
+            return;
+        }
+
+        this._client = new StoreApiClient();
+        const formData = FormSerializeUtil.serialize(form);
+
+        ElementLoadingIndicatorUtil.create(document.body);
+
+        const orderId = this.options.orderId;
+        const request = new XMLHttpRequest();
+        let callback = null;
+        if (orderId !== null) {
+            formData.set('orderId', orderId);
+            request.open('POST', '/sales-channel-api/v1/adyen/payment');
+            callback = this.afterSetPayment.bind(this);
+        } else {
+            request.open('POST', this.options.checkoutOrderUrl);
+            callback = this.afterCreateOrder.bind(this);
+        }
+        request.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+        request.setRequestHeader('sw-language-id', adyenCheckoutOptions.languageId);
+        this._client._sendRequest(request, formData, callback);
+    }
+
+    afterCreateOrder(response) {
+
+    }
+
+    afterSetPayment(response) {
+
+    }
+
+    afterPayOrder(response) {
+
+    }
+
+}
--- a/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
@@ -14,7 +14,7 @@ export default class ConfirmOrderPlugin extends Plugin {
     }
 
     confirmOrder(event) {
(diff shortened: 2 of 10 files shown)


src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+            request.open('POST', this.options.checkoutOrderUrl);
+            callback = this.afterCreateOrder.bind(this);
+        }
+        request.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+        request.setRequestHeader('sw-language-id', adyenCheckoutOptions.languageId);
+        this._client._sendRequest(request, formData, callback);
+    }
+
+    afterCreateOrder(response) {
```
cyattilakiss, line 44: Great job! 

Using this example I think we should hook on this event - we should create the order at first and do the initial payments call
In case the payment is refused, error or authorised we don't even need to disturb the flow everyhting can work as is

In case the payment is identifyShopper or challengeSHopper we need to return the action and create the component from action

Maybe a paymentStatus endpoint shuld be exposed which returns the first status after the order is created and we can use the paymentDetails endpoint if there is an extra necessary step needed.

src/Controller/SalesChannelApiController.php
```diff
@@ -128,4 +128,20 @@ public function postPaymentDetails(SalesChannelContext $context): JsonResponse
...
+     * @Route(
+     *     "/sales-channel-api/v1/adyen/payment",
+     *     name="sales-channel-api.action.adyen.payment",
+     *     methods={"POST"}
+     * )
+     *
+     * @param SalesChannelContext $context
+     * @return JsonResponse
+     */
+    public function payment(SalesChannelContext $context): JsonResponse
```
cyattilakiss, line 143:
```suggestion
    public function getPaymentStatus(SalesChannelContext $context): JsonResponse
```

src/Controller/SalesChannelApiController.php
```diff
@@ -128,4 +128,20 @@ public function postPaymentDetails(SalesChannelContext $context): JsonResponse
     {
         return new JsonResponse($this->paymentDetailsService->doPaymentDetails($context));
     }
+
+    /**
+     * @RouteScope(scopes={"sales-channel-api"})
+     * @Route(
+     *     "/sales-channel-api/v1/adyen/payment",
+     *     name="sales-channel-api.action.adyen.payment",
+     *     methods={"POST"}
```
cyattilakiss, line 137:
```suggestion
     *     "/sales-channel-api/v1/adyen/payment-status",
     *     name="sales-channel-api.action.adyen.payment-status",
     *     methods={"GET"}
```

src/Controller/SalesChannelApiController.php
```diff
@@ -128,4 +128,20 @@ public function postPaymentDetails(SalesChannelContext $context): JsonResponse
...
+     * )
+     *
+     * @param SalesChannelContext $context
+     * @return JsonResponse
+     */
+    public function payment(SalesChannelContext $context): JsonResponse
+    {
+        return new JsonResponse(['aloha']);
```
cyattilakiss, line 145: Create new pull in ticket to actually return the payment status

src/Subscriber/CheckoutSubscriber.php
```diff
@@ -0,0 +1,114 @@
...
+    public function onCheckoutConfirmLoaded(CheckoutConfirmPageLoadedEvent $event)
+    {
+        $salesChannelContext = $event->getSalesChannelContext();
+        $page = $event->getPage();
+        $page->addExtension(
+            self::ADYEN_DATA_EXTENSION_ID,
+            new ArrayEntity(
+                [
+                    'paymentUrl' => $this->router->generate(
+                        'sales-channel-api.action.adyen.payment',
```
cyattilakiss, line 102:
```suggestion
                        'sales-channel-api.action.adyen.payment-status',
```

src/Subscriber/CheckoutSubscriber.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function onCheckoutConfirmLoaded(CheckoutConfirmPageLoadedEvent $event)
+    {
+        $salesChannelContext = $event->getSalesChannelContext();
+        $page = $event->getPage();
+        $page->addExtension(
+            self::ADYEN_DATA_EXTENSION_ID,
+            new ArrayEntity(
+                [
+                    'paymentUrl' => $this->router->generate(
```
cyattilakiss, line 101:
```suggestion
                    'paymentStatusUrl' => $this->router->generate(
```

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -0,0 +1,64 @@
...
+/* eslint-disable no-unused-vars */
+export default class ConfirmOrderPlugin extends Plugin {
+
+    init() {
+        const confirmOrderForm = DomAccess.querySelector(document, '#confirmOrderForm');
+        confirmOrderForm.addEventListener('submit', this.confirmOrder.bind(this));
+    }
+
+    confirmOrder(event) {
+        event.preventDefault();
```
cyattilakiss, line 17: Some defensive programming here because I faced the issue when these vars were not available and then the whole process just died.

```suggestion
        if (!!adyenCheckoutOptions && !!adyenCheckoutOptions.paymentUrl && adyenCheckoutOptions.checkoutOrderUrl) {
        event.preventDefault();
```

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -0,0 +1,64 @@
...
+            formData.set('orderId', orderId);
+            request.open('POST', adyenCheckoutOptions.paymentUrl);
+            callback = this.afterSetPayment.bind(this);
+        } else {
+            request.open('POST', adyenCheckoutOptions.checkoutOrderUrl);
+            callback = this.afterCreateOrder.bind(this);
+        }
+        request.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+        request.setRequestHeader('sw-language-id', adyenCheckoutOptions.languageId);
+        this._client._sendRequest(request, formData, callback);
```
cyattilakiss, line 41:
```suggestion
        this._client._sendRequest(request, formData, callback);
        }
```

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -0,0 +1,64 @@
...
+
+        this._client = new StoreApiClient();
+        const formData = FormSerializeUtil.serialize(form);
+
+        ElementLoadingIndicatorUtil.create(document.body);
+
+        const orderId = this.options.orderId;
+        const request = new XMLHttpRequest();
+        let callback = null;
+        if (typeof(orderId) !== 'undefined' && orderId !== null) { //Only used if the order is being edited
```
cyattilakiss, line 31: I believe the same can be achieved with this but please double check with @KadoBOT 
```suggestion
        if (!!orderId) { //Only used if the order is being edited
```

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -0,0 +1,64 @@
...
+            request.open('POST', adyenCheckoutOptions.checkoutOrderUrl);
+            callback = this.afterCreateOrder.bind(this);
+        }
+        request.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+        request.setRequestHeader('sw-language-id', adyenCheckoutOptions.languageId);
+        this._client._sendRequest(request, formData, callback);
+    }
+
+    afterCreateOrder(response) {
+        const order = JSON.parse(response);
```
cyattilakiss, line 45:
```suggestion
        try {
            const order = JSON.parse(response);
        } catch () {
            // Response is not a valid JSON
            // TODO error handling
            return;
        }
```

src/Resources/views/storefront/page/checkout/confirm/confirm-payment.html.twig
```diff
@@ -0,0 +1,16 @@
...
+    {% if adyenFrontendData and adyenFrontendData.paymentUrl %}
+        <div id="adyen-checkout-options"
+             data-language-id="{{ adyenFrontendData.languageId }}"
+             data-payment-url="{{ adyenFrontendData.paymentUrl }}"
+             data-checkout-order-url="{{ adyenFrontendData.checkoutOrderUrl }}">
+        </div>
+    {% endif %}
+    <script>
+        var adyenCheckoutOptions = document.querySelector('#adyen-checkout-options').dataset;
+    </script>
```
cyattilakiss, line 15: Without the div, the script won't work so maybe it's better to render the script only along with the div
```suggestion
        <script>
            var adyenCheckoutOptions = document.querySelector('#adyen-checkout-options').dataset;
        </script>
    {% endif %}
```

src/Controller/SalesChannelApiController.php
acampos1916: PW-2768

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
KadoBOT: yep! Attila's suggestion is the same. Unless `orderId` `false` is a valid option, which I don't think it is :P

cyattilakiss: Thanks! 👍

PR #22, opened 2020-07-29
[PW-2689] 3DS1 & 3DS2 payments - Handle the backend

cyattilakiss:
## Summary
Handle the endpoint accepting the state.data and do the paymentDetails request


src/Handlers/PaymentResponseHandler.php
```diff
@@ -57,101 +69,208 @@ class PaymentResponseHandler
...
+    public function handleShopwareAPIs(
+        AsyncPaymentTransactionStruct $transaction,
+        SalesChannelContext $salesChannelContext,
+        PaymentResponseHandlerResult $paymentResponseHandlerResult
+    ) {
+        $orderTransactionId = $transaction->getOrderTransaction()->getId();
+        $resultCode = $paymentResponseHandlerResult->getResultCode();
+        $context = $salesChannelContext->getContext();
+
+        // Only store psp reference for the transaction if this is the firs/original pspreference
```
acampos1916, line 193:
```suggestion
        // Only store psp reference for the transaction if this is the first/original pspreference
```

src/Handlers/PaymentResponseHandler.php
```diff
@@ -67,120 +88,197 @@ public function __construct(
...
                 // Unsupported resultCode
                 $this->logger->error(
-                    "There was an error with the payment method. id:  " . $orderTransactionId .
+                    "There was an error with the payment method. id:  " .
                     ' Unsupported result code in response: ' . print_r($response, true)
                 );
+        }
+    }
+
+    public function handleShopwareAPIs(
```
acampos1916, line 175: Just for consistency with the second function
```suggestion
    public function handleShopwareApis(
```

src/Handlers/CardsPaymentMethodHandler.php
```diff
@@ -203,7 +203,11 @@ public function pay(
             );
         }
 
-        return $this->paymentResponseHandler->handlePaymentResponse($response, $transaction, $salesChannelContext);
+        $orderNumber = $transaction->getOrder()->getOrderNumber();
+
+        $result = $this->paymentResponseHandler->handlePaymentResponse($response, $orderNumber, $salesChannelContext);
+
+        return $this->paymentResponseHandler->handleShopwareAPIs($transaction, $salesChannelContext, $result);
```
acampos1916, line 210: Just for consistency with the second function
```suggestion
        return $this->paymentResponseHandler->handleShopwareApis($transaction, $salesChannelContext, $result);
```

cyattilakiss:
@msilvagarcia we are going to merge this one for now because @acampos1916 and I also depend on this PR for our current tickets. Could you please review it even after it's merged and flag anything you think should be changed? I can do that in a separate PR then.
I hope this can work for you as well :)

PR #25, opened 2020-07-30
[PW-2691] Handle response back from an issuer page

cyattilakiss:
## Summary
Whenever a shopper is redirected back from an issuer page (can be iDeal or can be 3DS1 etc) do a paymentDetails request validating the details coming back from the redirect page and handle the response from the paymentDetails using the same handler as we used before at the /payments request.

Some fixed for pay() and paymentResultHandler

## Tested scenarios
Mocked result request
Whole flow works until payment-status request


src/Handlers/ResultHandler.php
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
msilvagarcia, line 3:
```suggestion
<?php declare(strict_types=1);
```

cyattilakiss:
> In [src/Handlers/ResultHandler.php](https://github.com/Adyen/adyen-shopware6/pull/25/files#diff-2c4e35b326f8df7d7df55ca0f1ab949cR104), would be better to throw an exception from our module instead of one from Shopware core.
> 
> But it looks great as it is right now.

indeed, let me change that quickly.

PR #28, opened 2020-08-03
[PW-2678] Set applicationInfo data 

cyattilakiss:
## Summary
Add applicationInfo support
Add getModuleVersion() to retrieve the Adyen module version from the
composer.json file if available
Add ContainerParametersService.php as a helper service to retrieve
parameters from the container

## Tested scenarios
Available version in composer.json
Not available version in composer.json

### DIFF
--- a/src/Service/ClientService.php
+++ b/src/Service/ClientService.php
@@ -96,6 +96,7 @@ class ClientService extends \Adyen\Client
 
     /**
      * Get adyen module's version from composer.json
+     * TODO: switch to a shopware service to retrieve the module version instead of composer
      *
      * @return string
      */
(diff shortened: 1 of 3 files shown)


src/Service/ClientService.php
```diff
@@ -62,17 +82,45 @@ public function __construct(
...
-
             //TODO set $this->configuration
         } catch (\Exception $e) {
             $this->genericLogger->error($e->getMessage());
             // TODO: check if $environment is test and, if so, exit with error message
         }
     }
+
+    /**
+     * Get adyen module's version from composer.json
```
cyattilakiss, line 98:
```suggestion
     * Get adyen module's version from composer.json
     * TODO: switch to a shopware service to retrieve the module version instead of composer
```

PR #33, opened 2020-08-06
[PW-2821] Send the CSRF token to finalize-transaction

cyattilakiss:
## Summary
Add new endpoint to process redirect results from Adyen /adyen/redirect-result
Replace the original returnUrl with the new endpoint


src/Handlers/CardsPaymentMethodHandler.php
```diff
@@ -481,4 +500,37 @@ public function preparePaymentsRequest(
...
+    {
+        // Parse the original return URL to retrieve the query parameters
+        $returnUrlQuery = parse_url($returnUrl, PHP_URL_QUERY);
+
+        // In case the URL is malformed it cannot be parsed
+        if (false === $returnUrlQuery) {
+            throw new PaymentException('Return URL is malformed');
+        }
+
+        // Generate the custome Adyen endpoint to receive the redirect from the issuer page
```
acampos1916, line 522:
```suggestion
        // Generate the custom Adyen endpoint to receive the redirect from the issuer page
```

PR #45, opened 2020-08-20
[PW-2788] Add payment methods to the plugin and adjust card payment handler

acampos1916:
## Summary
New payment methods have been added to be installed and picked up by the generic component in the checkout page. In order to outsource the payment handler logic an abstract class was created along with a static function to fetch the PM code in the Payment Subscriber using the handler.

All the payment methods have been added to services.xml with all the classes' dependencies, maybe there's a way to simplify those declarations.

New frontend validation must be implemented to accommodate all the payment methods.

Some API endpoints were moved from v2 to v3 since that version was removed from S6.3.

## Tested scenarios
New payment methods are installed and rendered in the frontend.

**Fixed issue**:  PW-2788

### DIFF
--- a/src/Handlers/KlarnaPaymentMethodHandler.php
+++ b/src/Handlers/KlarnaPaymentMethodHandler.php
@@ -39,20 +39,4 @@ class KlarnaPaymentMethodHandler extends AbstractPaymentMethodHandler implements
     {
         return 'klarna';
     }
-
-    public function pay(
-        AsyncPaymentTransactionStruct $transaction,
-        RequestDataBag $dataBag,
-        SalesChannelContext $salesChannelContext
-    ): RedirectResponse {
-        return parent::pay($transaction, $dataBag, $salesChannelContext);
-    }
-
-    public function finalize(
-        AsyncPaymentTransactionStruct $transaction,
-        Request $request,
-        SalesChannelContext $salesChannelContext
-    ): void {
-        parent::finalize($transaction, $request, $salesChannelContext);
-    }
 }
--- a/src/Handlers/KlarnaPaymentMethodHandler.php
+++ b/src/Handlers/KlarnaPaymentMethodHandler.php
@@ -35,7 +35,8 @@ use Symfony\Component\HttpFoundation\Request;
 class KlarnaPaymentMethodHandler extends AbstractPaymentMethodHandler implements AsynchronousPaymentHandlerInterface
 {
 
-    static function getPaymentMethodCode(){
+    public static function getPaymentMethodCode()
+    {
         return 'klarna';
     }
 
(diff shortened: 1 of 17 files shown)


msilvagarcia (commented):
For implementations of the AbstractPaymentMethodHandler, you don't need to implement the `pay` and `finalize` methods if they're just calling `parent::$method` :)

src/Handlers/KlarnaPaymentMethodHandler.php
```diff
@@ -0,0 +1,42 @@
...
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+class KlarnaPaymentMethodHandler extends AbstractPaymentMethodHandler implements AsynchronousPaymentHandlerInterface
+{
+
+    public static function getPaymentMethodCode()
+    {
+        return 'klarna';
```
cyattilakiss, line 40: I'm not sure if this is going to work - there are multiple Klarna method codes in Adyen - when we implement open invoices we can start testing this

src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -158,7 +166,8 @@ export default class CheckoutPlugin extends Plugin {
     }
 
     onConfirmPayment (event) {
-        if (!this.cardFormValidator.validateForm()) {
+        //TODO Implement validation for multiple PMs
+        if (!this.cardFormValidator.validateForm() && false) {
```
cyattilakiss, line 170: I think just the name should be changed since the component does the validation for us with a common validate function

cyattilakiss (approved):

src/Handlers/KlarnaPaymentMethodHandler.php
acampos1916: Yes, we can check for the correct one. I found a few that would apply and was not sure which one was needed.

src/Resources/app/storefront/src/checkout/checkout.plugin.js
acampos1916: Do you think we should do it in this PR? I had created a ticket for that (PW-2914) but we could do it here

cyattilakiss: it's fine in the other ticket :) I will resolve this comment

src/Handlers/KlarnaPaymentMethodHandler.php
cyattilakiss: please create a ticket for this to be sure we won't forget :)

acampos1916: PW-2922

PR #48, opened 2020-09-03
[PW-2786] OpenInvoice lines support and Klarna PMs

acampos1916:
## Summary
This PR adds order item lines to the payments request in order to support open invoice PMs. It also includes Klarna Pay Later and Pay Now.

The services.xml abstract payment method handler declaration now is parent to the concrete handlers and some old use statements were removed.

## Tested scenarios
Payments with Klarna Pay Later and Pay Now.
Payments with 3DS and non 3DS cards.

**Fixed issue**:  PW-2786

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -54,6 +54,8 @@ use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 abstract class AbstractPaymentMethodHandler
 {
 
+    static protected $isOpenInvoice = false;
+
     /**
      * @var CheckoutService
      */
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -64,6 +64,8 @@ use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 abstract class AbstractPaymentMethodHandler
 {
 
+    const PROMOTION = 'promotion';
+
     protected static $isOpenInvoice = false;
 
     /**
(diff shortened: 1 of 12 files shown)


msilvagarcia (approved):

src/PaymentMethods/KlarnaPayNowPaymentMethod.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return string
+     */
+    public function getDescription(): string
+    {
+        return ''; //TODO
```
cyattilakiss, line 48: do we need to add any description here? Is there maybe something from our docs we can copy - @acampos1916 can you please check with Daniel maybe?

src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -331,6 +365,13 @@ public function preparePaymentsRequest(
         //Validate state.data for payment and build request object
         $request = $this->checkoutStateDataValidator->getValidatedAdditionalData($request);
 
+        //Setting payment method type if not present in statedata
+        if (empty($request['paymentMethod']['type'])) {
```
cyattilakiss, line 369: when was it empty?

src/Handlers/ResultHandler.php
```diff
@@ -115,15 +115,13 @@ public function processResult(
...
-            if (empty($md) || empty($paRes)) {
-                throw new PaymentException('MD and/or PaRes parameter is missing from the redirect request');
-            }
+            $redirectResult = $request->query->get('redirectResult');
 
             // Construct the details object for the paymentDetails request
             $details = [
                 'MD' => $md,
-                'PaRes' => $paRes
+                'PaRes' => $paRes,
```
cyattilakiss, line 123: I think it would be better to only add to the details when the variable is not empty like:

```
if (!empty($paRes)) {
    $details['PaRes'] = $paRes;
}
```

src/Handlers/ResultHandler.php
```diff
@@ -115,15 +115,13 @@ public function processResult(
             // Get MD and PaRes to be validated
             $md = $request->query->get('MD');
             $paRes = $request->query->get('PaRes');
-
-            if (empty($md) || empty($paRes)) {
-                throw new PaymentException('MD and/or PaRes parameter is missing from the redirect request');
-            }
+            $redirectResult = $request->query->get('redirectResult');
```
cyattilakiss, line 118: constants can be created for these then:

MD
PaRes
redirectResult

If these are not existing consts already :)

cyattilakiss:
some phpcs is still failing as well

src/PaymentMethods/KlarnaPayNowPaymentMethod.php
acampos1916: Ok, I'm checking with him, I'll create a separate PR for the description of all PMs

src/Handlers/AbstractPaymentMethodHandler.php
acampos1916: When the PM doesn't create a state.data with the generic component, so in that case the request starts as an empty array https://github.com/Adyen/adyen-shopware6/blob/develop/src/Handlers/AbstractPaymentMethodHandler.php#L315

src/Handlers/ResultHandler.php
acampos1916: Added the constants 👍

src/Handlers/ResultHandler.php
acampos1916: Done!

cyattilakiss (approved):

PR #55, opened 2020-09-08
[PW-2787] New paypal integration

cyattilakiss:
## Summary
Add paypal support in checkout without using the adyen checkout components

Known issue, will be fixed by shopware in the core: https://issues.shopware.com/issues/NEXT-8636
When the customer cancels the payment the payment status will be failed in the admin and the customer will see a message related to failed payment and not cancelled payment.

## Tested scenarios
paypal successful payment
paypal cancelled payment

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -693,7 +693,6 @@ abstract class AbstractPaymentMethodHandler
 
         switch ($exceptionType) {
             case PaymentCancelledException::class:
-                $this->logger->debug('ATTILA CANCELLED DEBUG');
                 throw new CustomerCanceledAsyncPaymentException(
                     $transactionId,
                     $exception->getMessage()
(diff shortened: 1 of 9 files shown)


src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -686,4 +682,32 @@ private function getProduct(string $productId, Context $context): ProductEntity
...
+            case PaymentFailedException::class:
+                throw new AsyncPaymentFinalizeException(
+                    $transactionId,
+                    $exception->getMessage()
+                );
+            default:
+                throw new AsyncPaymentFinalizeException(
+                    $transactionId,
+                    $exception->getMessage()
+                );
```
acampos1916, line 710: Should we add a break; and combine the second alternative with default? Needs testing though 
```suggestion
                break;
            case PaymentFailedException::class:
            default:
                throw new AsyncPaymentFinalizeException(
                    $transactionId,
                    $exception->getMessage()
                );
```

cyattilakiss: Looks good to me

acampos1916 (dismissed):

msilvagarcia (changes requested):
I do not agree on reusing code for handling exception. This is part of our business logic and later, if we need to adapt something for either `finalize` or `pay`, we will start adding a lot of complexity when it should be decoupled to begin with instead.

We could use [named constructors for exceptions](https://www.alainschlesser.com/structuring-php-exceptions/) instead of handling all exceptions in a private method.

msilvagarcia (approved):

acampos1916 (approved):

PR #56, opened 2020-09-09
[PW-3004] Fix payment when attempt without state.data stored but requested from db

cyattilakiss:
## Summary
A 500 error can occur 
 - when the customer doesn't complete the component before saving the payment method configuration therefore the state.data is missing from the /payments request because of the missing frontend validation
 - when the customer arrives to the payment methods list page while their preferred payment method is already selected therefore they don't change and store it but places the order right away

## Tested scenarios

Wrong details filled in the checkout component - validation won't let the customer to save the changes anymore
Payment method change is not saved but order is placed - redirects the user to the finalize page and shows an error message while the payment is failed in the admin page

### DIFF
--- a/src/Resources/app/storefront/src/checkout/checkout.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/checkout.plugin.js
@@ -24,7 +24,7 @@ import DomAccess from 'src/helper/dom-access.helper';
 import Plugin from 'src/plugin-system/plugin.class';
 import StoreApiClient from 'src/service/store-api-client.service';
 import validatePaymentMethod from '../validator/paymentMethod';
-import CardFormValidator from '../validator/CardFormValidator';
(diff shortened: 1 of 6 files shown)


acampos1916 (approved):

src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -169,8 +166,11 @@ export default class CheckoutPlugin extends Plugin {
     }
 
     onConfirmPayment (event) {
-        //TODO Implement validation for multiple PMs
-        if (!this.cardFormValidator.validateForm() && false) {
+        if (!(this.getSelectedPaymentMethodHandlerIdentifyer() in this.formValidator)) {
```
msilvagarcia, line 169: Nice use of brackets to emphasize what's part of negation

msilvagarcia (approved):

PR #73, opened 2020-10-14
[PW-2842] Getting module version from plugin entity

### LINKED ISSUE #49: [PW-2842] Install with composer fails due to missing/not required autoload.php
**Describe the bug**

If trying to install the plugin using the project's composer there is an error upon activating / refreshing the plugin list as it tries to require a missing `autoload.php` which is not necessary as the dependencies are already loaded by the project's main `autoload.php`.

**To Reproduce**
Add 
```
    "repositories": [
        {
            "type": "git",
            "url": "https://github.com/Adyen/adyen-shopware6.git"
        }
    ],
```

to the project's `composer.json`.

Execute: `composer require "adyen/adyen-shopware6:dev-develop"`

Execute 
```
bin/console plugin:refresh
bin/console plugin:install --clearCache --activate AdyenPaymentShopware6
```

**Expected behavior**
In the file for `\Adyen\Shopware\AdyenPaymentShopware6` the `require_once __DIR__ . '/../vendor/autoload.php';` should only be done, if not installed by composer. Probably the easiest way is to only require the file if it exists.

acampos1916:
## Summary
The plugin used to get the module version from the composer file and this caused problems with the directory structure. We're switching to the plugin entity to fetch this value now, leveraging the cache system.

## Tested scenarios
Shopware 6.3.2.0
`getModuleVersion()` queries the plugin entity to get the installed version. In production mode this value is then saved in cache for future use.

**Fixed issue**:  PW-2842 Fixes #49

### DIFF
--- a/src/AdyenPaymentShopware6.php
+++ b/src/AdyenPaymentShopware6.php
@@ -140,4 +140,6 @@ class AdyenPaymentShopware6 extends Plugin
     }
 }
 
-require_once __DIR__ . '/../vendor/autoload.php';
+if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
+    require_once __DIR__ . '/../vendor/autoload.php';
+}
--- a/src/Service/ClientService.php
+++ b/src/Service/ClientService.php
@@ -130,7 +130,8 @@ class ClientService extends Client
     }
 
     /**
-     * Get adyen module's version from PluginEntity
+     * Get adyen module's version from cache if exists or from PluginEntity
+     * Stores the module version in the cache when empty
      *
      * @return string
      */
--- a/src/Service/ClientService.php
+++ b/src/Service/ClientService.php
@@ -26,12 +26,19 @@ namespace Adyen\Shopware\Service;
 
 use Adyen\Client;
 use Psr\Log\LoggerInterface;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepositoryInterface;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\Plugin\PluginEntity;
 use Shopware\Core\Framework\Store\Services\StoreService;
+use Psr\Cache\CacheItemPoolInterface;
 
 class ClientService extends Client
 {
     const MERCHANT_APPLICATION_NAME = 'adyen-shopware6';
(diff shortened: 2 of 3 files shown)


src/AdyenPaymentShopware6.php
```diff
@@ -140,4 +140,6 @@ private function setPaymentMethodIsActive(bool $active, Context $context, string
     }
 }
 
-require_once __DIR__ . '/../vendor/autoload.php';
+if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
```
cyattilakiss, line 143: what happens if the file doesn't exist? We need the modules from composer otherwise we will get php warnings at other parts of the code - shouldn't we throw an exception here?

src/Service/ClientService.php
```diff
@@ -108,33 +130,37 @@ public function getClient($salesChannelId)
     }
 
     /**
-     * Get adyen module's version from composer.json
-     * TODO: switch to a shopware service to retrieve the module version instead of composer
+     * Get adyen module's version from PluginEntity
```
cyattilakiss, line 133:
```suggestion
     * Get adyen module's version from cache if exists or from PluginEntity
     * Stores the module version in the cache when empty
```

src/Service/ClientService.php
```diff
@@ -108,33 +130,37 @@ public function getClient($salesChannelId)
...
         }
 
-        $composerJson = json_decode($composerJson, true);
-        if (json_last_error() !== JSON_ERROR_NONE) {
-            $this->genericLogger->error('composer.json is not a valid JSON in the Adyen plugin folder');
-            return "NA";
+        //Prefer the cached module version
+        if (!$moduleVersionCacheItem->isHit() || !$moduleVersionCacheItem->get()) {
+            /** @var PluginEntity $pluginEntity */
+            $pluginEntity = $this->pluginRepository->search(
```
cyattilakiss, line 148: this is always the installed version right? Is this something we need to keep an eye during testing when we deploy different versions of the code but we don't install it again?

src/AdyenPaymentShopware6.php
acampos1916: What I had in mind was that if the autoload file is not there it is because the plugin was installed through composer like described here https://github.com/Adyen/adyen-shopware6/issues/49#issue-692793542 in which case the dependencies are loaded with the Shopware autoload. What do you think?

src/Service/ClientService.php
acampos1916: Yes this is always the installed version. I don't think it will bring issues with testing as this is only used for the `merchantApplication` field in API requests

cyattilakiss (approved):

msilvagarcia (approved):

PR #75, opened 2020-10-16
[PW-2922] Fix payment method code for Klarna

msilvagarcia:
## Summary

Klarna uses code `klarna_account` instead of just `klarna`.

## Tested scenarios

None

**Fixed issue**:


cyattilakiss (changes requested):
sent you a message in mm

PR #77, opened 2020-10-19
[PW-2839] Removing plugin data on uninstall

acampos1916:
## Summary
If during uninstallation the user selects "Remove all plugin data permanently" the configuration keys used by the bundle are deleted and the plugin's tables are dropped.

Payment methods are not removed as the IDs could still be referenced from order_transaction.

## Tested scenarios
Shopware 6.3.2.0
Uninstall with "Remove all plugin data permanently" checked: Deactivates all payment methods, deletes configuration keys and drops adyen tables.
Uninstall with "Remove all plugin data permanently" unchecked: Deactivates all payment methods.

**Fixed issue**:  PW-2839

### DIFF
--- a/src/AdyenPaymentShopware6.php
+++ b/src/AdyenPaymentShopware6.php
@@ -25,7 +25,6 @@
 
 namespace Adyen\Shopware;
 
-use Adyen\Shopware\Handlers\CardsPaymentMethodHandler;
 use Adyen\Shopware\PaymentMethods\PaymentMethods;
 use Adyen\Shopware\PaymentMethods\PaymentMethodInterface;
 use Shopware\Core\Framework\Plugin;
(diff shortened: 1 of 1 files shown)


cyattilakiss (dismissed):

src/AdyenPaymentShopware6.php
```diff
@@ -52,23 +57,57 @@ public function install(InstallContext $installContext): void
...
     {
+        parent::uninstall($uninstallContext);
+
+        //Deactivating payment methods
         foreach (PaymentMethods::PAYMENT_METHODS as $paymentMethod) {
-            $this->setPaymentMethodIsActive(false, $uninstallContext->getContext(), $paymentMethod);
+            $this->setPaymentMethodIsActive(false, $uninstallContext->getContext(), new $paymentMethod());
+        }
+
+        //Call parent function and exit if the user prefers to keep the plugin's data
```
acampos1916, line 82:
```suggestion
        //Exit here if the user prefers to keep the plugin's data
```

cyattilakiss (approved):

msilvagarcia (approved):

PR #79, opened 2020-10-20
[PW-3182] New 3DS1 same site cookie policy fix

cyattilakiss:
## Tested scenarios

**Fixed issue**:


src/Migration/Migration1603384890AdyenPaymentResponse.php
```diff
@@ -0,0 +1,58 @@
...
+                inner join `order` o on apr.order_number = o.order_number
+                inner join `order_transaction` ot on o.id = ot.order_id and o.version_id = ot.order_version_id
+            set apr.order_transaction_id = ot.id
+            where 1=1;
+        ");
+        $connection->executeUpdate("
+            alter table `adyen_payment_response`
+                drop column order_number,
+                drop column sales_channel_api_context_token;
+        ");
```
acampos1916, line 32: Should we move this update to updateDestructive since it is dropping columns?

src/Migration/Migration1603384890AdyenPaymentResponse.php
```diff
@@ -0,0 +1,58 @@
...
+            alter table `adyen_payment_response`
+                add column order_number varchar(64) after id,
+                add column sales_channel_api_context_token varchar(255) after order_number;        ");
+        $connection->executeUpdate("
+            update `adyen_payment_response` apr
+                inner join `order` o on apr.order_number = o.order_number
+                inner join `order_transaction` ot on o.id = ot.order_id and o.version_id = ot.order_version_id
+            set apr.order_number = o.order_number
+            where 1=1;
+        ");
```
acampos1916, line 47: And this one to update()?

src/Migration/Migration1603384890AdyenPaymentResponse.php
msilvagarcia: If I understood correctly, `update` is called when upgrading the plugin and `upgradeDestructive` is called when downgrading it.

PR #83, opened 2020-11-03
[PW-2789 ] Enable save card details

msilvagarcia:
## Summary

This adds support for saving card details. A future pull request will enable paying with stored credit cards.

The data is stored in Adyen servers securely, and only a token is shared between the plugin and Adyen.

## Tested scenarios

Payment with credit cards, asking to save the card details or not.

**Fixed issue**:

### DIFF
--- a/src/Resources/app/storefront/src/checkout/checkout.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/checkout.plugin.js
@@ -144,6 +144,10 @@ export default class CheckoutPlugin extends Plugin {
             onChange: this.onPaymentMethodChange.bind(this)
         });
 
+        if (paymentMethod.type === 'scheme') {
+            configuration.enableStoreDetails = true;
+        }
+
         try {
             const paymentMethodInstance = window.adyenCheckout
                 .create(paymentMethod.type, configuration);
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -369,6 +369,11 @@ abstract class AbstractPaymentMethodHandler
             $paymentMethodType = $request['paymentMethod']['type'];
         }
 
+        if ($request['storePaymentMethod'] === true) {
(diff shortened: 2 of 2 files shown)


src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -144,6 +144,10 @@ export default class CheckoutPlugin extends Plugin {
             onChange: this.onPaymentMethodChange.bind(this)
         });
 
+        if (paymentMethod.type === 'scheme') {
+            configuration.enableStoreDetails = true;
```
acampos1916, line 148: I'm guessing we should enable this only if the checkbox is ticked? Part of the OneClick payments ticket

acampos1916 (approved):

src/Resources/app/storefront/src/checkout/checkout.plugin.js
cyattilakiss: We should only enable this when the user is logged in and for now the payment method is scheme
Later we are going to add more payment methods to the list which can be tokenised.

This configuration serves the purpose to show the checkbox.

acampos1916: Ah didn't see this was part of the configuration before creating the component 👍

msilvagarcia: @cyattilakiss we have a user ID even if the user is not signed in in Shopware. So if the session lasts for long enough, the token can be reused.

cyattilakiss: I would only let them see the checkbox if they are logged in otherwise it can be confusing that we let them store the payment method but then they are probably never able to reuse the stored payment method. Also I don't think it's secure to rely on the session in this case because there is no way for the customer to break their sessions (like logging out) and they would left their stored payment method in the session. The second part is pretty though I was just thinking loud :)

cyattilakiss (approved):

PR #84, opened 2020-11-05
[PW-2461] OneClick payment methods

acampos1916:
## Summary
After saving the payment details for future use this PR enables the payment with saved methods.

## Tested scenarios
Payments with stored card details, 3DS1/3DS2, optional and required CVC

**Fixed issue**:  PW-2461


src/Handlers/OneClickPaymentMethodHandler.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Adyen\Shopware\Handlers;
+
+use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\AsynchronousPaymentHandlerInterface;
+
+class OneClickPaymentMethodHandler extends AbstractPaymentMethodHandler implements AsynchronousPaymentHandlerInterface
+{
+
+    public static function getPaymentMethodCode()
+    {
+        return 'storedPaymentMethods';
```
cyattilakiss, line 35: can this conflict with recurring stored payment methods in the future if we add recurring as well?

src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -166,6 +171,54 @@ export default class CheckoutPlugin extends Plugin {
         }
     }
 
+    renderStoredPaymentMethod(paymentMethod) {
+        //  if the container doesn't exits don't try to render the component
+        const paymentMethodContainer = $('[data-adyen-payment-method="' + this.paymentMethodTypeHandlers["oneclick"] + '"]');
+
+        // container doesn't exist, something went wrong on the template side
+        // If payment method doesn't have details, just skip it
```
cyattilakiss, line 179: Not true anymore, comment can be removed
```suggestion
```

src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -166,6 +171,54 @@ export default class CheckoutPlugin extends Plugin {
...
+        });
+
+        /*Use the storedPaymentMethod object and the custom onChange function as the configuration object together*/
+        const configuration = Object.assign(paymentMethod, {
+            onChange: this.onPaymentMethodChange.bind(this)
+        });
+
+        if (paymentMethod.type === 'scheme') {
+            configuration.enableStoreDetails = true;
+        }
```
cyattilakiss, line 202: Is it needed for oneclick payment methods?
I think it's only required for payment methods where you do the oneoff payment and using that payment you create the token at Adyen. Therefore for already stored payment methods I believe this config is not needed.

```suggestion
```

src/Handlers/OneClickPaymentMethodHandler.php
acampos1916: We can create a Recurring payment method handler that uses the same code if necessary, as long as only OneClick shows in the checkout. It wouldn't be a conflict.

cyattilakiss: sounds good :)

msilvagarcia (approved):

src/Subscriber/PaymentSubscriber.php
```diff
@@ -215,12 +218,11 @@ public function onCheckoutConfirmLoaded(PageLoadedEvent $event)
...
                     ),
-                    'paymentMethodsResponse' => json_encode(
-                        $this->paymentMethodsService->getPaymentMethods($salesChannelContext)
-                    ),
+                    'paymentMethodsResponse' => json_encode($paymentMethodsResponse),
                     'orderId' => $orderId,
                     'stateDataPaymentMethod' => $stateDataPaymentMethod,
-                    'pluginId' => $adyenPluginId
+                    'pluginId' => $adyenPluginId,
+                    'storedPaymentMethods' => $paymentMethodsResponse['storedPaymentMethods'] ?? []
```
cyattilakiss, line 225: don't you need to json_encode this as well?

cyattilakiss (approved):

PR #93, opened 2020-12-07
[PW-3610] Implementing client key

acampos1916:
## Summary
Adyen can use the client key to authenticate requests from the payment environment. We implement a new config field for client key and if present it is used it instead of fetching the origin key. Origin key is kept as a fallback.

Generic component bundle was updated to 3.12.1.

## Tested scenarios
Rendering the component with and without a client key config.

**Fixed issue**:  PW-3610


src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -61,10 +61,11 @@ export default class CheckoutPlugin extends Plugin {
             );
         }
 
-        const { locale, originKey, environment, paymentMethodsResponse } = adyenCheckoutConfiguration;
+        const { locale, originKey, clientKey, environment, paymentMethodsResponse } = adyenCheckoutConfiguration;
         const ADYEN_CHECKOUT_CONFIG = {
             locale,
             originKey,
+            clientKey,
```
cyattilakiss, line 68: Can the component handle all the scenarios?

1. Origin key is provided, Client key is empty string
2. Origin key is provided, Client key provided (which one is preferred by the component?)
3. Origin key is empty string, Client key is provided

Maybe it's better to add the clientKey to the ADYEN_CHECKOUT_CONFIG if provided otherwise the origin key?

src/Resources/config/config.xml
```diff
@@ -40,6 +40,22 @@
             <name>hmacLive</name>
             <label>HMAC Key for live</label>
         </input-field>
+        <input-field type="password">
+            <name>clientKeyTest</name>
+            <label>Client Key for test</label>
+            <helpText>
+                We use your client key to authenticate requests from your payment environment.
+                See docs.adyen.com for more information.
```
cyattilakiss, line 48: can you provide a more specific docs link?

src/Resources/config/config.xml
```diff
@@ -40,6 +40,22 @@
...
+            </helpText>
+        </input-field>
+        <input-field type="password">
+            <name>clientKeyLive</name>
+            <label>Client Key for live</label>
+            <helpText>
+                We use your client key to authenticate requests from your payment environment.
+                See docs.adyen.com for more information.
```
cyattilakiss, line 56: here as well please

cyattilakiss (commented):
Could you also tag the origin key only function as deprecated and will be removed in the version 2.0.0?

src/Resources/app/storefront/src/checkout/checkout.plugin.js
acampos1916: Yes, I actually tested this and is looking good:

1. Origin key is provided, Client key is empty string: Origin key is used successfully
2. Origin key is provided, Client key provided (which one is preferred by the component?): [Client key is used successfully](https://github.com/Adyen/adyen-shopware6/pull/93/files#diff-94c3ec56b7a92d80e1953086cde504fc4919edaefee871ac714fdf3fd3e57562R214) 
3. Origin key is empty string, Client key is provided: Client key is used successfully.

cyattilakiss: Nice thanks for the confirmation 👍

src/Resources/config/config.xml
acampos1916: Changed it to the specific link and suggested the new Client Key section to docs.

src/Resources/config/config.xml
acampos1916: Changed it to the specific link and suggested the new Client Key section to docs.

acampos1916:
> Could you also tag the origin key only function as deprecated and will be removed in the version 2.0.0?

Added!

cyattilakiss (approved):
Looks great! 🥇

msilvagarcia (approved):

cyattilakiss (approved):

PR #98, opened 2020-12-08
Add Bancontact card payment method

cyattilakiss:
## Summary

Add the Bancontact component into Shopware 6

## Tested scenarios

Pay with bancontact component
Update card details on bancontact payment method

**Fixed issue**:


src/Service/PaymentStateDataService.php
```diff
@@ -130,7 +132,15 @@ public function getPaymentMethodType(string $contextToken): ?string
             return OneClickPaymentMethodHandler::getPaymentMethodCode();
         }
 
-        return $stateDataArray['paymentMethod']['type'];
+        $type = $stateDataArray['paymentMethod']['type'];
```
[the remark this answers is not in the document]
acampos1916: Yes, sounds like it should be possible to use the payment method handler instead of `$stateDataArray['paymentMethod']['type']`. We would have to check for the stored payment methods and bcmc handlers to add the exceptions. What do you think @peterojo?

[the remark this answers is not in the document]
acampos1916: I think with this approach the exceptions won't be needed as bcmc and stored PMs would be checked against the handler in the frontend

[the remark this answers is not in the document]
peterojo: I think it should work, we already maintain a mapping in `adyenConfiguration.paymentMethodTypeHandlers` that we would need to convert the `handler_adyen_...` -style formatted identifiers into the short method names

src/Resources/app/storefront/src/checkout/checkout.plugin.js
```diff
@@ -182,14 +182,17 @@ export default class CheckoutPlugin extends Plugin {
...
-            element.find('[data-adyen-payment-container]').hide();
-            element.find('[data-adyen-update-payment-details]').show();
-        } else if (method === 'oneclick' && adyenCheckoutOptions.statedataPaymentMethod === 'storedPaymentMethods') {
-            //The state data stored matches storedPaymentMethods, show update details button for oneclick
-            $(`[data-adyen-payment-method=${adyenConfiguration.paymentMethodTypeHandlers["oneclick"]}]`).find('[data-adyen-payment-container]').hide();
-            $(`[data-adyen-payment-method=${adyenConfiguration.paymentMethodTypeHandlers["oneclick"]}]`).find('[data-adyen-update-payment-details]').show();
+        let selectedPaymentMethod = this.getSelectedPaymentMethodKey();
+        if (method === selectedPaymentMethod) {
+            if (method == 'oneclick') {
+                //The state data stored matches storedPaymentMethods, show update details button for oneclick
```
peterojo, line 188: since we are no more returning `storedPaymentMethod` as the type for oneclick, the method and selectedPaymentMethod will match, like all the others.

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -35,7 +35,7 @@ export default class ConfirmOrderPlugin extends Plugin {
             const confirmPaymentModal = $('#confirmPaymentModal');
             // Do validation if the payment method has all the required data submitted by the customer
             // 1. case: State data is not stored
-            if (!adyenCheckoutOptions.statedataPaymentMethod) {
+            if (!adyenCheckoutOptions.stateDataIsStored) {
```
peterojo, line 38: We still needed a way to check if the state data is stored

peterojo (dismissed):

src/Subscriber/PaymentSubscriber.php
```diff
@@ -247,7 +247,7 @@ public function onCheckoutConfirmLoaded(PageLoadedEvent $event)
 
         $page->setPaymentMethods($filteredPaymentMethods);
 
-        $stateDataPaymentMethod = $this->paymentStateDataService->getPaymentMethodType(
+        $stateDataIsStored = (bool) $this->paymentStateDataService->getPaymentStateDataFromContextToken(
```
cyattilakiss, line 250: @peterojo just to double check, do we remove the state.data when we select a payment method without stat.data after we've just selected and submitted one method with state.data?

peterojo: Yes we do, see [here](https://github.com/Adyen/adyen-shopware6/blob/61af19439f5670f5c7977ff703242ee551ea97e8/src/Subscriber/PaymentSubscriber.php#L431). When saving a new PM, if the state is empty, we delete any existing state data for the context

cyattilakiss: Thanks!

src/Service/PaymentStateDataService.php
```diff
@@ -109,30 +109,6 @@ public function getPaymentStateDataFromContextToken(string $contextToken): ?Paym
         return $stateDataRow;
     }
 
-    /**
-     * @param string $contextToken
-     * @return string|null
-     */
-    public function getPaymentMethodType(string $contextToken): ?string
```
acampos1916, line 116: Being a public function should we deprecate it until next major release instead?

acampos1916 (dismissed):

src/Service/PaymentStateDataService.php
peterojo: Good point

peterojo (approved):

acampos1916 (approved):

PR #99, opened 2020-12-09
[PW-3631] storedPaymentMethods state.data check

acampos1916:
## Summary
The payment method state.data checker that triggers the modal to show was preventing all payments with Klarna because of a handler identifier mismatch.

Additionally, the storedPaymentMethods state.data is valid with no user input, which causes it to be saved when no-input PMs were selected (e.g. Klarna). A new check was introduced in PaymentSubscriber.php to verify that the state.data and the PM selected match, if they don't match then the state.data is deleted as it could come from a previous selection.

## Tested scenarios
Payments with Klarna now don't trigger the modal.
Payments with Klarna don't use the storedPaymentMethods state.data.

**Fixed issue**:  PW-3631


src/Subscriber/PaymentSubscriber.php
```diff
@@ -130,15 +142,41 @@ public static function getSubscribedEvents(): array
...
+            $selectedPaymentMethodIsStoredPM =
+                $selectedPaymentMethod->getFormattedHandlerIdentifier() == 'handler_adyen_oneclickpaymentmethodhandler';
+
+            $stateDataIsStoredPM = !empty($stateDataArray["paymentMethod"]["storedPaymentMethodId"]);
+
+            //Only store the state data if it matches the selected PM
+            if ($stateDataIsStoredPM == $selectedPaymentMethodIsStoredPM) {
+                try {
+                    $this->paymentStateDataService->insertPaymentStateData(
+                        $event->getSalesChannelContext()->getToken(),
```
cyattilakiss, line 168: Does it maybe a good idea to also insert the selected payment method (id, handler, code - one of these which identifies the payment method in shopware) so later we don't have to rely on the satte data type?

acampos1916: I think so, it could be handy, just not on this subscriber as we always check the new oncoming state.data against the selected PM. Created a ticket to look into it PW-3633

src/Subscriber/PaymentSubscriber.php
```diff
@@ -130,15 +142,41 @@ public static function getSubscribedEvents(): array
...
-                $this->paymentStateDataService->insertPaymentStateData(
-                    $event->getSalesChannelContext()->getToken(),
-                    $event->getRequestDataBag()->get('adyenStateData'),
-                    $event->getRequestDataBag()->get('adyenOrigin')
+        //State data from the frontend
+        $stateData = $event->getRequestDataBag()->get('adyenStateData');
+
+        if ($stateData) {
+            //Convert the state data into an array
+            $stateDataArray = json_decode($stateData, true);
```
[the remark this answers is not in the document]
acampos1916: Right, added this JSON error check

src/Subscriber/PaymentSubscriber.php
```diff
@@ -130,15 +142,41 @@ public static function getSubscribedEvents(): array
...
+            //Only store the state data if it matches the selected PM
+            if ($stateDataIsStoredPM == $selectedPaymentMethodIsStoredPM) {
+                try {
+                    $this->paymentStateDataService->insertPaymentStateData(
+                        $event->getSalesChannelContext()->getToken(),
+                        $event->getRequestDataBag()->get('adyenStateData'),
+                        $event->getRequestDataBag()->get('adyenOrigin')
+                    );
+                } catch (AdyenException $exception) {
+                    //TODO handle error
```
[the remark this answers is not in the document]
acampos1916: Done with an error log and error message on the frontend

cyattilakiss (approved):

src/Resources/app/storefront/src/configuration/adyen.js
```diff
@@ -25,9 +25,9 @@ export default {
     paymentMethodTypeHandlers: {
         'scheme': 'handler_adyen_cardspaymentmethodhandler',
         'ideal': 'handler_adyen_idealpaymentmethodhandler',
-        'klarna': 'handler_adyen_klarnaaccountpaymentmethodhandler',
-        'klarna_account': 'handler_adyen_klarnapaynowpaymentmethodhandler',
-        'klarna_paynow': 'handler_adyen_klarnapaylaterpaynowpaymentmethodhandler',
+        'klarna': 'handler_adyen_klarnapaylaterpaymentmethodhandler',
+        'klarna_account': 'handler_adyen_klarnaaccountpaymentmethodhandler',
+        'klarna_paynow': 'handler_adyen_klarnapaynowpaymentmethodhandler',
```
msilvagarcia, line 30: Is this required for `state.data`?

acampos1916: Not specifically for `state.data`. It is needed to validate the payment methods that require input before submitting the order here https://github.com/Adyen/adyen-shopware6/blob/develop/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js#L37

peterojo (approved):

src/Resources/app/storefront/src/configuration/adyen.js
msilvagarcia: I mean this change. This does not seem strictly related to the pull request

acampos1916: Correct, this is a bugfix necessary for the storedPaymentMethods state.data check, the PR description has both explanations. Do you think we should move it to a separate PR?

msilvagarcia (approved):

PR #101, opened 2020-12-14
[PW-3591] add method in NotificationService to set scheduled_processing_time on Notification

peterojo:
## Summary
Add method in `NotificationService` to set the scheduled processing time on a notification, which will be triggered via the scheduled task (shopware cron job) for every notification that needs to be scheduled.

## Tested scenarios

`scheduled_processing_time` field is updated in database.
the whole flow can be tested when the scheduled task is implemented in a subsequent PR

**Fixed issue**:  PW-3591

### DIFF
--- a/src/Service/NotificationReceiverService.php
+++ b/src/Service/NotificationReceiverService.php
@@ -90,6 +90,7 @@ class NotificationReceiverService
     public function process(Request $requestObject)
     {
         $request = $requestObject->request->all();
+        $salesChannelId = $requestObject->attributes->get('sw-sales-channel-id');
         $basicAuthUser = $requestObject->server->get('PHP_AUTH_USER');
         $basicAuthPassword = $requestObject->server->get('PHP_AUTH_PW');
 
@@ -123,7 +124,7 @@ class NotificationReceiverService
             // Process each notification item
             foreach ($request['notificationItems'] as $notificationItem) {
                 $notificationItem['NotificationRequestItem']['live'] = $request['live'];
-                if (!$this->processNotificationItem($notificationItem['NotificationRequestItem'])) {
+                if (!$this->processNotificationItem($notificationItem['NotificationRequestItem'], $salesChannelId)) {
                     throw new ValidationException();
                 }
             }
@@ -250,16 +251,17 @@ class NotificationReceiverService
      * Save notification into the database for cron job to execute notification
      *
      * @param $notificationItem
+     * @param $salesChannelId
      * @return bool
      * @throws AdyenException
      * @throws AuthenticationException
      * @throws HMACKeyValidationException
      * @throws MerchantAccountCodeException
      */
-    protected function processNotificationItem($notificationItem)
+    protected function processNotificationItem($notificationItem, $salesChannelId)
     {
         $merchantAccount = $this->configurationService->getMerchantAccount();
-        $hmacKey = $this->configurationService->getHmacKey();
+        $hmacKey = $this->configurationService->getHmacKey($salesChannelId);
 
         // validate the notification
         if ($this->isValidated($notificationItem, $merchantAccount, $hmacKey)) {
@@ -268,12 +270,10 @@ class NotificationReceiverService
                 print_r($notificationItem, true));
 
             // skip report notifications
-            {
             if ($this->isReportNotification($notificationItem['eventCode'])) {
                 $this->logger->info('Notification is a REPORT notification from ' .
                     'Adyen Customer Area');
                 return true;
-            }
             }
 
             // check if notification already exists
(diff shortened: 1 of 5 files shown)


src/Service/NotificationReceiverService.php
```diff
@@ -90,6 +90,7 @@ public function __construct(
     public function process(Request $requestObject)
     {
         $request = $requestObject->request->all();
+        $salesChannelId = $requestObject->attributes->get('sw-sales-channel-id');
```
acampos1916, line 93: If I'm not mistaken this is executed with no context, so there wouldn’t be a sales channel ID. Or is the Headless sales channel used?

peterojo: I believe the headless sales channel is used, I can get that the salesChannelId by inspecting the request parameterBag

src/Service/NotificationReceiverService.php
```diff
@@ -250,16 +251,17 @@ private function isAuthorized($isTestNotification, $requestUser, $requestPasswor
...
      * @throws AuthenticationException
      * @throws HMACKeyValidationException
      * @throws MerchantAccountCodeException
      */
-    protected function processNotificationItem($notificationItem)
+    protected function processNotificationItem($notificationItem, $salesChannelId)
     {
         $merchantAccount = $this->configurationService->getMerchantAccount();
-        $hmacKey = $this->configurationService->getHmacKey();
+        $hmacKey = $this->configurationService->getHmacKey($salesChannelId);
```
acampos1916, line 264: If the sales channel ID is empty I think it fetches the default value ("All Sales Channel"), if not then it will always get the Headless sales channel ID, can you confirm that?

Maybe setting the notifications HMAC in Headless is the way to go, let's talk about it later

src/Service/NotificationService.php
```diff
@@ -72,43 +74,56 @@ public function insertNotification(array $notification): void
...
         }
 
         $this->notificationRepository->create(
             [$fields],
-            \Shopware\Core\Framework\Context::createDefaultContext()
+            Context::createDefaultContext()
+        );
+    }
+
+    public function setNotificationSchedule(string $notificationId, DateTimeInterface $scheduledProcessingTime)
```
peterojo, line 117: went for this function name to distinguish this method from the `scheduleNotification` method that we will add in the scheduled task.

acampos1916 (approved):

src/Service/NotificationReceiverService.php
```diff
@@ -122,7 +123,8 @@ public function process(Request $requestObject)
 
             // Process each notification item
             foreach ($request['notificationItems'] as $notificationItem) {
-                if (!$this->processNotificationItem($notificationItem['NotificationRequestItem'])) {
+                $notificationItem['NotificationRequestItem']['live'] = $request['live'];
```
cyattilakiss, line 126: why is this set here?

src/Service/NotificationReceiverService.php
cyattilakiss: Yesterday I think we came to the conclusion to ask the merchants (via the documentation) to only set the HMAC for the default store configuration and only use this config from there and not for each store separately. Is this still needed here?

peterojo: We built the notification endpoint ((see here)[https://github.com/Adyen/adyen-shopware6/blob/develop/src/Storefront/Controller/NotificationReceiverController.php#L55]) to use the `storefront`scope and, from my investigation, the notification request comes in via the `storefront` salesChannelID and that is what will be used here to fetch the HMAC key.
Moreover, I think it is still needed because in the (`getHmacKey` function)[https://github.com/Adyen/adyen-shopware6/blob/develop/src/Service/ConfigurationService.php#L164] we still use the sales channel to query for the Hmac key, not passing the sales channel will return a php error.

src/Service/NotificationReceiverService.php
peterojo: This was one of the bugs found when testing, `live` is needed when [creating the `NotificationEntity`](https://github.com/Adyen/adyen-shopware6/blob/develop/src/Service/NotificationService.php#L102) and it is not set inside the `NotificationRequestItem` object of the request ([see request format in docs](https://docs.adyen.com/development-resources/webhooks/understand-notifications?tab=codeBlockOU4K1_1#notification-structure))

src/Service/NotificationReceiverService.php
cyattilakiss: got it :)

peterojo:
@cyattilakiss I can make another PR if you have follow up comments, I realized that I merged this PR too quickly

PR #102, opened 2020-12-16
[PW-3589] Add getScheduledUnprocessedNotifications() to NotificationService

peterojo:
## Summary

Retrieve a collection of notifications:
 * between T-1 day and T-1 minute, 
 * that are not done and 
 * not being processed and 
 * `scheduled_processing_time` is not null, 
 * limited to 100 rows at a time, 
 * sorted by `scheduled_processing_time`

## Tested scenarios

Returns collection that fits above criteria

**Fixed issue**:  PW-3589

### DIFF
--- a/src/Service/NotificationReceiverService.php
+++ b/src/Service/NotificationReceiverService.php
@@ -119,7 +119,7 @@ class NotificationReceiverService
         $pluginIsLive = $this->configurationService->getEnvironment() === Environment::LIVE;
 
         // Is the notification coming from live environment?
-        if(!is_bool($request['live'] ?? null)) {
+        if (!is_bool($request['live'] ?? null)) {
             $request['live'] = isset($request['live']) && $request['live'] === 'true';
         }
 
--- a/src/Service/NotificationReceiverService.php
+++ b/src/Service/NotificationReceiverService.php
@@ -115,10 +115,11 @@ class NotificationReceiverService
         }
 
         $acceptedMessage = '[accepted]';
+        $isLive = isset($request['live']) && $request['live'] === 'true';
 
         // Process each notification item
         foreach ($request['notificationItems'] as $notificationItem) {
-            $notificationItem['NotificationRequestItem']['live'] = isset($request['live']) && $request['live'] === 'true';
+            $notificationItem['NotificationRequestItem']['live'] = $isLive;
             if (!$this->processNotificationItem($notificationItem['NotificationRequestItem'], $salesChannelId)) {
                 throw new ValidationException();
             }
--- a/src/Service/NotificationReceiverService.php
+++ b/src/Service/NotificationReceiverService.php
@@ -24,6 +24,7 @@
 
 namespace Adyen\Shopware\Service;
 
+use Adyen\Environment;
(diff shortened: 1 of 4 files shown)


src/Service/NotificationReceiverService.php
```diff
@@ -114,35 +115,16 @@ public function process(Request $requestObject)
...
+        if(!is_bool($request['live'] ?? null)) {
+            $request['live'] = isset($request['live']) && $request['live'] === 'true';
+        }
 
-            return new JsonResponse(
-                $this->returnAccepted($acceptedMessage)
-            );
-        } else {
+        // Checks if notification mode and the store mode configuration matches
+        if ($request['live'] !== $pluginIsLive) {
```
acampos1916, line 127: Maybe we don't need to reject notifications if the modes don't match. What do you think @cyattilakiss?

cyattilakiss: @acampos1916 We can receive and store those notifications - we should not do anything else other than the security checks before storing them in the db. 
When we process them though we should log an error and stop processing the notification which has the mismatch.

peterojo: Okay, I will remove this particular line of validation

acampos1916: Gotcha, so yes @peterojo, we can remove the mode check from the receiving service

acampos1916 (approved):

src/Service/NotificationService.php
```diff
@@ -74,37 +78,37 @@ public function insertNotification(array $notification): void
...
+        if (array_key_exists('success', $notification)) {
             $fields['success'] = "true" === $notification['success'];
         }
-        if (!empty($notification['paymentMethod'])) {
+        if (array_key_exists('paymentMethod', $notification)) {
             $fields['paymentMethod'] = $notification['paymentMethod'];
         }
-        if (!empty($notification['amount']['value'])) {
-            $fields['amountValue'] = (string)$notification['amount']['value'];
+        if (array_key_exists('value', $notification['amount'] ?? [])) {
```
cyattilakiss, line 99: isn't it going to throw an exception if the `$notification['amount']` does not exist?
Maybe better to use isset() instead of array_key_exists()
(isset though will return false if the value is null)

peterojo: I used the null coalesce operator (`?? []`) so that it doesn't throw an exception, this way it will use the empty array if the `amount` key is missing and then array_key_exists will return false. Because this is a nested array, with isset() I would have to do something like `if (isset($notification['amount']) && isset($notification['amount']['value']))`

cyattilakiss: @peterojo isset() handles multi level check as well and handles if the first key doesn't even exist so you could do only:
`isset($notification['amount']['value'])`

peterojo: You're right, I can change it all to `isset()`. I changed from `!empty()` because of `$notification['live']` which could have a value of `false` and `!empty()` would return false. But `isset()` also works correctly with this scenario

cyattilakiss (approved):

PR #111, opened 2020-12-29
PW-3120 add weekly cron to fetch PM logos

peterojo:
## Summary

Each payment method has a function: public function getLogo(): string to retrieve the logo but we return the https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/sofort.png' url instead of downloading it and providing it from the server of the merchant.
Show the logos using the resources from the server.

### Changes
Fix involves creation of scheduled task  to fetch logos weekly and upload them to create Shopware media entities and then attaching them to the payment_method entities.
Weekly cron will first delete the old media and then recreate new ones from the downloaded file
Fallback is that in the case where a remote URL is temporarily unreachable, the payment method will be skipped, thereby preserving its media from the previous run.
New Command `adyen:fetch-logos` is also introduced to fetch logos on demand.

### Requirement
Merchants need to make sure that the following configuration is set to enabled:
`shopware.media.enable_url_upload_feature`
This is enabled by default in `platform/src/Core/Framework/Resources/config/packages/shopware.yaml`

New MediaFolder `Adyen` will be created with migration.
## Tested scenarios

Logos are downloaded into Media entities and attached to PMs, display accordingly in checkout and in Administration content section.
Logos are overwritten with every run of scheduled task.
Unavailable remote? PM is skipped.

### DIFF
--- a/src/PaymentMethods/KlarnaPayNowPaymentMethod.php
+++ b/src/PaymentMethods/KlarnaPayNowPaymentMethod.php
@@ -85,7 +85,7 @@ class KlarnaPayNowPaymentMethod implements PaymentMethodInterface
      */
     public function getLogo(): string
     {
-        return 'klarna.png';
+        return 'klarna_paynow.png';
     }
 
     /**
--- a/src/PaymentMethods/KlarnaPayNowPaymentMethod.php
+++ b/src/PaymentMethods/KlarnaPayNowPaymentMethod.php
@@ -85,7 +85,7 @@ class KlarnaPayNowPaymentMethod implements PaymentMethodInterface
      */
     public function getLogo(): string
     {
-        return 'https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/klarna.png';
+        return 'klarna.png';
     }
 
     /**
--- a/src/PaymentMethods/OneClickPaymentMethod.php
+++ b/src/PaymentMethods/OneClickPaymentMethod.php
@@ -85,7 +85,7 @@ class OneClickPaymentMethod implements PaymentMethodInterface
      */
     public function getLogo(): string
     {
-        return 'https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/card.png';
+        return 'card.png';
     }
 
     /**
--- a/src/ScheduledTask/FetchPaymentMethodLogosHandler.php
+++ b/src/ScheduledTask/FetchPaymentMethodLogosHandler.php
@@ -30,6 +30,7 @@ use Adyen\Shopware\Service\ConfigurationService;
 use Exception;
 use Psr\Log\LoggerAwareTrait;
 use Shopware\Core\Checkout\Payment\PaymentMethodEntity;
+use Shopware\Core\Content\Media\File\MediaFile;
 use Shopware\Core\Content\Media\MediaService;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepositoryInterface;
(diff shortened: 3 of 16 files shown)


src/PaymentMethods/KlarnaPayNowPaymentMethod.php
```diff
@@ -85,7 +85,7 @@ public function getTemplate(): ?string
      */
     public function getLogo(): string
     {
-        return 'https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/klarna.png';
+        return 'klarna.png';
```
cyattilakiss, line 88:
```suggestion
        return 'klarna_paynow.png';
```

src/PaymentMethods/OneClickPaymentMethod.php
```diff
@@ -85,7 +85,7 @@ public function getTemplate(): ?string
      */
     public function getLogo(): string
     {
-        return 'https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/card.png';
+        return 'card.png';
```
cyattilakiss, line 88: Could you pls create a follow up ticket to download all the icons for all the card types available for the merchant like: visa, bcmc etc?
For now it's a perfect solution but since for the one click payment methods we know the card type we can also show the logo for that.

src/ScheduledTask/FetchPaymentMethodLogosHandler.php
```diff
@@ -0,0 +1,180 @@
...
+
+    public static function getHandledMessages(): iterable
+    {
+        return [ FetchPaymentMethodLogos::class ];
+    }
+
+    public function run(): void
+    {
+        if (!$this->enableUrlUploadFeature) {
+            $this->logger->debug('Configuration `shopware.media.enable_url_upload_feature` is disabled.');
```
cyattilakiss, line 91: good one +1
Is it a config that the merchant needs to enable to make sure they can use this?
Please add this as a step to the release note and as a suggestion to docs (let me know if you need help with that ;) )

src/ScheduledTask/FetchPaymentMethodLogosHandler.php
```diff
@@ -0,0 +1,180 @@
...
+
+            // Look up corresponding payment_method entity.
+            $result = $this->paymentMethodRepository->search(
+                (new Criteria())->addFilter(new EqualsFilter(
+                    'handlerIdentifier',
+                    $paymentMethod->getPaymentHandler()
+                )),
+                $context
+            );
+            if ($result->getTotal() === 0) {
```
cyattilakiss, line 109: Is it correct like this?
```suggestion

            // Skip if the payment method is not registered in ShopWare
            if ($result->getTotal() === 0) {
```

acampos1916 (commented):
@peterojo nice PR. Is there a way to trigger the logo update job on demand? Maybe some merchants would like to refresh the logos through a CLI command that runs `FetchPaymentMethodLogosHandler:run()`, or right after plugin install.

src/ScheduledTask/FetchPaymentMethodLogosHandler.php
peterojo: Yes

src/ScheduledTask/FetchPaymentMethodLogosHandler.php
peterojo: Usually it is enabled by default but yes they do need to make sure it it enabled if they want logos

peterojo:
> @peterojo nice PR. Is there a way to trigger the logo update job on demand? Maybe some merchants would like to refresh the logos through a CLI command that runs `FetchPaymentMethodLogosHandler:run()`, or right after plugin install.

@acampos1916 I think we could create a command to trigger this on demand. I will create a ticket for this

cyattilakiss (approved):

acampos1916 (dismissed):

cyattilakiss (approved):

acampos1916 (approved):

peterojo:
I added the command to trigger the logos update, since it should be part of this ticket

PR #115, opened 2021-01-04
[PW-2843] implement notification auth from the library

peterojo:
## Summary

Use: https://github.com/Adyen/adyen-php-api-library/blob/develop/src/Adyen/Service/NotificationReceiver.php services instead of maintaining the same code in shopware6 here: https://github.com/Adyen/adyen-shopware6/blob/develop/src/Service/NotificationReceiverService.php
## Tested scenarios

Receiving of notifications still works the same:
- Authentication checks, Hmac checks still work as expected
- Test notifications still return expected results

### DIFF
--- a/src/Service/NotificationReceiverService.php
+++ b/src/Service/NotificationReceiverService.php
@@ -105,10 +105,10 @@ class NotificationReceiverService
 
         // Authorize notification
         if (!$this->notificationReceiver->isAuthenticated(
-                $firstNotificationItem,
-                $this->configurationService->getMerchantAccount(),
-                $basicAuthUser,
-                $basicAuthPassword
+            $firstNotificationItem,
+            $this->configurationService->getMerchantAccount(),
+            $basicAuthUser,
+            $basicAuthPassword
         )) {
             throw new AuthenticationException();
         }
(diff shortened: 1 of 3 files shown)


src/Service/NotificationReceiverService.php
```diff
@@ -95,23 +94,23 @@ public function process(Request $requestObject)
...
-        $isTestNotification = $this->isTestNotification($request);
+        // First item in the notification
+        $firstNotificationItem = $request['notificationItems'][0]['NotificationRequestItem'];
 
         // Authorize notification
-        if (!$this->isAuthorized($isTestNotification, $basicAuthUser, $basicAuthPassword)) {
-            throw new AuthorizationException();
+        if (!$this->notificationReceiver->isAuthenticated(
+            $firstNotificationItem,
+            $this->configurationService->getMerchantAccount(),
```
acampos1916, line 109: Should we start using the sales channel ID from the request object (`sw-sales-channel` I believe) to get the merchant account and not only the default one?

acampos1916 (changes requested):

src/Service/NotificationReceiverService.php
peterojo: @acampos1916 sure we can do that, we already have the variable in this method. Could that make a difference with this authentication?

acampos1916: Ah yes, let's use that variable `$salesChannelId`. I also see that the `isAuthenticated()` function should receive the config user and pass and not the `$_SERVER` vars as the library method already does that. Maybe that's an old bug.

acampos1916 (approved):

cyattilakiss (approved):

PR #117, opened 2021-01-07
[PW-3694] add notification processors

peterojo:
## Summary

Process notifications
Handle AUTHORISATION and OFFER_CLOSED notifications according to the following logic.
Screen Shot 2021-01-08 at 10 44 29 AM
Handle edge case: In case the last payment method used for the order is a non adyen payment method then don't process the notifications
I created a few tickets for further improvements. See comments in PW-3694
## Tested scenarios

Payment states are changed accordingly when scheduled task runs

**Fixed issue**:

### DIFF
--- a/src/Migration/Migration1607529255AdyenNotification.php
+++ b/src/Migration/Migration1607529255AdyenNotification.php
@@ -15,8 +15,8 @@ class Migration1607529255AdyenNotification extends MigrationStep
     public function update(Connection $connection): void
     {
         $connection->executeUpdate("
-            alter table `adyen_notification`
-            add column scheduled_processing_time datetime null after processing;
+            ALTER TABLE `adyen_notification`
+            ADD COLUMN `scheduled_processing_time` DATETIME(3) NULL AFTER `processing` COMMENT 'Scheduled for';
         ");
     }
 
--- a/src/NotificationProcessor/AuthorisationNotificationProcessor.php
+++ b/src/NotificationProcessor/AuthorisationNotificationProcessor.php
@@ -34,16 +34,25 @@ class AuthorisationNotificationProcessor extends NotificationProcessor implement
         $orderTransaction = $this->getOrder()->getTransactions()->first();
         $state = $orderTransaction->getStateMachineState()->getTechnicalName();
         $context = Context::createDefaultContext();
+        $logContext = [
+            'orderId' => $this->getOrder()->getId(),
+            'orderNumber' => $this->getOrder()->getOrderNumber(),
+            'eventCode' => NotificationEventCodes::AUTHORISATION,
+            'originalState' => $state
+        ];
 
-        $paymentMethod = $orderTransaction->getPaymentMethod();
         if ($this->getNotification()->isSuccess()) {
             if ($state !== OrderTransactionStates::STATE_PAID) {
                 $this->getTransactionStateHandler()->paid($orderTransaction->getId(), $context);
+                $logContext['newState'] = OrderTransactionStates::STATE_PAID;
             }
         } else {
             if ($state === OrderTransactionStates::STATE_IN_PROGRESS) {
                 $this->getTransactionStateHandler()->fail($orderTransaction->getId(), $context);
+                $logContext['newState'] = OrderTransactionStates::STATE_FAILED;
             }
         }
+
+        $this->logger->info('Processed ' . NotificationEventCodes::AUTHORISATION . ' notification.', $logContext);
     }
 }
--- a/src/NotificationProcessor/AuthorisationNotificationProcessor.php
+++ b/src/NotificationProcessor/AuthorisationNotificationProcessor.php
@@ -27,7 +27,7 @@ namespace Adyen\Shopware\NotificationProcessor;
 use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionStates;
 use Shopware\Core\Framework\Context;
 
-class OfferClosedProcessor extends BaseProcessor
+class AuthorisationNotificationProcessor extends BaseNotificationProcessor
 {
     public function process()
     {
@@ -35,8 +35,15 @@ class OfferClosedProcessor extends BaseProcessor
         $state = $orderTransaction->getStateMachineState()->getTechnicalName();
         $context = Context::createDefaultContext();
 
-        if ($this->getNotification()->isSuccess() && $state === OrderTransactionStates::STATE_IN_PROGRESS) {
-            $this->getTransactionStateHandler()->fail($orderTransaction->getId(), $context);
+        $paymentMethod = $orderTransaction->getPaymentMethod();
+        if ($this->getNotification()->isSuccess()) {
+            if ($state !== OrderTransactionStates::STATE_PAID) {
+                $this->getTransactionStateHandler()->paid($orderTransaction->getId(), $context);
+            }
+        } else {
+            if ($state === OrderTransactionStates::STATE_IN_PROGRESS) {
+                $this->getTransactionStateHandler()->fail($orderTransaction->getId(), $context);
+            }
         }
     }
 }
--- a/src/NotificationProcessor/AuthorisationNotificationProcessor.php
+++ b/src/NotificationProcessor/AuthorisationNotificationProcessor.php
@@ -29,7 +29,7 @@ use Shopware\Core\Framework\Context;
 
 class AuthorisationNotificationProcessor extends NotificationProcessor implements NotificationProcessorInterface
 {
-    public function process()
+    public function process(): void
     {
         $orderTransaction = $this->getOrder()->getTransactions()->first();
         $state = $orderTransaction->getStateMachineState()->getTechnicalName();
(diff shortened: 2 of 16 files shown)


src/NotificationProcessor/AuthorisationProcessor.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\NotificationProcessor;
+
+use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionStates;
+use Shopware\Core\Framework\Context;
+
+class AuthorisationProcessor extends BaseProcessor
```
cyattilakiss, line 30: what do you think about: 
```suggestion
class AuthorisationNotificationProcessor extends BaseProcessor
```

The reason because if we use is somewhere we might not see the namespace and it can be confusing what is authorisationProcessor

src/NotificationProcessor/BaseProcessor.php
```diff
@@ -0,0 +1,98 @@
...
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2021 Adyen B.V.
```
cyattilakiss, line 18: 🎉 2021

src/NotificationProcessor/BaseProcessor.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * @var NotificationEntity
+     */
+    protected $notification;
+    /**
+     * @var OrderTransactionStateHandler
+     */
+    protected $transactionStateHandler;
+
+    public function process()
```
cyattilakiss, line 47: is it needed? Should this may be in an interface instead?

src/NotificationProcessor/NotificationEvents.php
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) 2021 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\NotificationProcessor;
+
+final class NotificationEvents
```
cyattilakiss, line 27: Maybe a bit more clear:
```suggestion
final class NotificationEventCodes
```

src/NotificationProcessor/NotificationEvents.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public const EVENT_AUTHORISATION = 'AUTHORISATION';
```
cyattilakiss, line 29: And if we change the class name we can remove the EVENT_ from the code consts since the class name will be visible always
```suggestion
    public const AUTHORISATION = 'AUTHORISATION';
```

src/NotificationProcessor/OfferClosedProcessor.php
```diff
@@ -0,0 +1,42 @@
...
+namespace Adyen\Shopware\NotificationProcessor;
+
+use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionStates;
+use Shopware\Core\Framework\Context;
+
+class OfferClosedProcessor extends BaseProcessor
+{
+    public function process()
+    {
+        $orderTransaction = $this->getOrder()->getTransactions()->first();
```
cyattilakiss, line 34: can there be more than 1 transaction in case of multiple payment attempt?

peterojo: Yes, there can. In the case an order fails and then succeeds, the order still has multiple transactions, but I did a few data dumps and I found that the OTs are listed in reverse chronological order.

src/NotificationProcessor/AuthorisationProcessor.php
peterojo: Sounds good to me 👍🏿  I was caught in between 2 minds on some of these class names :)

src/NotificationProcessor/BaseProcessor.php
peterojo: it's in the interface, that's why I have to implement it here.

cyattilakiss: I think you can move the interface implementation to the final classes and the abstract class doesn't need to have it implemented

src/NotificationProcessor/OfferClosedProcessor.php
cyattilakiss: Could you please create a ticket with low prio to discuss the possibility of processing each notification while connecting them to a transaction?

I don't know if there is any added value but maybe based on pspreference we can connect the transaction with the notification and therefore we only need to process the notifications which are connected to the last transaction.
I am not sure though if we have pspreference for each attempt - I think for redirectShoppers it's not necessary to have one.

Let's discuss it under a new ticket but probably it won't change anything :)

src/NotificationProcessor/BaseProcessor.php
peterojo: Ah I see your idea, I should convert this class to an abstract class and create another concrete class that will serve as the generic notification processor (for all the other event codes) and then only the concrete classes will implement the interface and extend this abstract class.

src/Migration/Migration1607529255AdyenNotification.php
```diff
@@ -15,8 +15,8 @@ public function getCreationTimestamp(): int
     public function update(Connection $connection): void
     {
         $connection->executeUpdate("
-            alter table `adyen_notification`
-            add column scheduled_processing_time datetime null after processing;
+            ALTER TABLE `adyen_notification`
+            ADD COLUMN `scheduled_processing_time` DATETIME(3) NULL AFTER `processing` COMMENT 'Scheduled for';
```
acampos1916, line 19: Why `DATETIME(3)`? And would this be applied in instances where this migration was already applied? Or should we move it to a new migration?

acampos1916 (commented):
Looks good! See comment about the migration

src/Migration/Migration1607529255AdyenNotification.php
peterojo: I made this optional change in a separate commit because it is only about being consistent with the other migration `\Adyen\Shopware\Migration\Migration1589892467AdyenNotification` which creates the table.
I can only still change this because this branch `PW-3049` is not yet merged into `develop` and therefore this is still a new migration. If you have already run this migration in your VM then no it will not change anything.

acampos1916: Ah that's right, this hasn't been released yet 😂

acampos1916 (approved):

src/NotificationProcessor/AuthorisationNotificationProcessor.php
```diff
@@ -0,0 +1,49 @@
...
+    public function process(): void
+    {
+        $orderTransaction = $this->getOrder()->getTransactions()->first();
+        $state = $orderTransaction->getStateMachineState()->getTechnicalName();
+        $context = Context::createDefaultContext();
+
+        $paymentMethod = $orderTransaction->getPaymentMethod();
+        if ($this->getNotification()->isSuccess()) {
+            if ($state !== OrderTransactionStates::STATE_PAID) {
+                $this->getTransactionStateHandler()->paid($orderTransaction->getId(), $context);
```
cyattilakiss, line 41: can you pls add some logs that indicates that the state has been changed because of the notification? (Also for all the other state changes)

src/NotificationProcessor/GenericNotificationProcessor.php
```diff
@@ -0,0 +1,35 @@
...
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\NotificationProcessor;
+
+class GenericNotificationProcessor extends NotificationProcessor implements NotificationProcessorInterface
+{
+    public function process(): void
+    {
+        // For other notification event codes that are not listed
```
cyattilakiss, line 31: please also add a log here that the notification has been processed but no action since it's not implemented yet in the plugin

src/NotificationProcessor/NotificationProcessorFactory.php
```diff
@@ -0,0 +1,54 @@
...
+        NotificationEventCodes::OFFER_CLOSED => OfferClosedNotificationProcessor::class,
+    ];
+
+    public static function create(
+        NotificationEntity $notification,
+        OrderEntity $order,
+        OrderTransactionStateHandler $transactionStateHandler
+    ): NotificationProcessorInterface {
+        /** @var NotificationProcessor $notificationProcessor */
+        $notificationProcessor = array_key_exists($notification->getEventCode(), self::$adyenEventCodeProcessors)
```
cyattilakiss, line 44: Very nice approach!

cyattilakiss (approved):

cyattilakiss (approved):

PR #119, opened 2021-01-11
[PW-3049] Add support for processing notifications

peterojo:
## Summary

Implement a service to process the notifications and close / tag as payed the orders based on the resultCode.

* PW-2843 Implement notification auth from the library
* PW-3588 Add new field to shopware.adyen_notification db table
* PW-3587 Add scheduled tasks to run every minute
* PW-3589 Add getScheduledUnprocessedNotifications() to NotificationService
* PW-3590 Add getUnscheduledNotifications() to NotificationService
* PW-3591 Add scheduleNotifications($scheduledProcessingTime) to NotificationService to update the scheduled_processing_time
* PW-3592 Add scheduleNotifications() to schedule notification in a scheduled task
* PW-3694 Process notifications
## Tested scenarios

Payment statuses are updated according to notification events

**Fixed issue**:


src/Migration/Migration1607529255AdyenNotification.php
```diff
@@ -0,0 +1,26 @@
...
+    public function update(Connection $connection): void
+    {
+        $connection->executeUpdate("
+            ALTER TABLE `adyen_notification`
+            ADD COLUMN `scheduled_processing_time` DATETIME(3) NULL AFTER `processing` COMMENT 'Scheduled for';
+        ");
+    }
+
+    public function updateDestructive(Connection $connection): void
+    {
```
msilvagarcia, line 24: Add a drop column for the destructive update.

msilvagarcia (changes requested):
I really like the Notification Processor classes. It's a great idea to limit the scope of methods and reduce the complexity in files.

I did leave a comment in the migration file, though, asking to implement the destructive update method. :)

Keep it up, looking great!

src/Migration/Migration1607529255AdyenNotification.php
cyattilakiss: I read about migrations nowadays and I am not sure about it but I think the updateDestructive is not for the downgrading but for the commands that cannot be reversed.
https://docs.shopware.com/en/shopware-platform-dev-en/developer-guide/migrations#modifying-the-database

Let's try to validate it with SW6 and then maybe we need to change the previous migration scripts.

msilvagarcia: Oh! That actually makes a lot of sense! :)

src/ScheduledTask/ProcessNotificationsHandler.php
```diff
@@ -0,0 +1,197 @@
...
+
+    public function run(): void
+    {
+        $context = Context::createDefaultContext();
+        $notifications = $this->notificationService->getScheduledUnprocessedNotifications();
+
+        foreach ($notifications->getElements() as $notification) {
+            /** @var NotificationEntity $notification */
+
+            $this->markAsProcessing($notification->getId());
```
cyattilakiss, line 102: Not sure if this is going to be an issue or not just thinking about an edge case:

If for any reason the cron job is running multiple times then it might happen that we pick up the same 100 notifications to be processed. Then maybe just to be more fail safe we could check also in this foreach if the notification is in progress or is already done and then we can still skip it.

I can imagine later when we process more notification types the required time of processing 1 notification can go up and can happen that 1 minute won't be enough for the processing.

Please also think about it maybe there are other edge cases.

cyattilakiss (dismissed):

src/ScheduledTask/ProcessNotificationsHandler.php
peterojo: The `getScheduledUnprocessedNotifications` function will only pick up notifications that have `processing=0` AND `done=0` and it is sorted by `scheduled_processing_time`.
When we mark as processing, the notification already does not meet this criteria anymore, therefore it cannot be picked up again.
In the case of an Exception, when we requeue the notification by updating the `scheduled_processing_time`, that way it is picked up last after all the other notifications that are already in the table.

cyattilakiss: okay, let's keep it like this for now and if we see that this is happening that cron job is running multiple times then we can think about it again

cyattilakiss (dismissed):

acampos1916:
Cool! Let's remember to update the docs with the notifications/payment status flows when this is released.

acampos1916 (dismissed):

acampos1916 (approved):

cyattilakiss (approved):

Issue #127, opened 2021-01-26
[PW-3922] Problem with completing a failed payment.


kamelz:
**Describe the bug**
Whenever we have a failed payment with Credit Card then we try to pay it again but this time with PayPal or Klarna we get this error: [screenshot](https://storage.googleapis.com/mind-mshop-prod-media/support/adyen-bug.png)

We checked [PaymentDetailsService.php#164](https://github.com/Adyen/adyen-shopware6/blob/d30f1001dfeeaa977eb8db0c4f7bfe7dbb95a824/src/Service/PaymentDetailsService.php#L164) and we found a missing return statement there as you will see in the error.

When we checked the logs we found the following entries:
```
[2021-01-25 16:08:30] adyen_generic.ERROR: There was an error with the payment method. Result code "Error" in response: Array ( [pspReference] => 4556115909096866 [resultCode] => Error [merchantReference] => 10645 [paymentMethod] => paypal [shopperLocale] => de-DE ) [] []
[2021-01-25 16:11:46] adyen_generic.ERROR: There was an error with the payment method. Result code "Error" in response: Array ( [pspReference] => 4556115909096866 [resultCode] => Error [merchantReference] => 10645 [paymentMethod] => paypal [shopperLocale] => de-DE ) [] []
[2021-01-25 16:11:46] adyen_generic.ERROR: Field 'details' expected to be a Structure but is an Array [] []
[2021-01-25 16:17:19] adyen_generic.ERROR: The payment was refused, order transaction merchant reference: 10646 [] []
```
**To Reproduce**
1. With live mode enabled, make an order with a test card for example to get a failed payment.
2. As a logged in customer now, go to "My orders" in your account, and in your order history, you'll see the failed order.
3. Click on "Complete order" to go again to the checkout page.
4. use PayPal or Klarna as a payment method(the first time it happened to us with PayPal then we tried it again with Klarna and we saw the same error).
5.Finish order and pay then you should see the error in the screenshot.
 
**Expected behavior**
To go to PayPal or Klarna payment page or in case of failer go back to the shop with an error message.  

**Additional context**
We use version 1.4.0.

acampos1916:
Hello @kamelz, thanks for opening this issue.

I was able to reproduce this and we have created an internal ticket to look into it. We'll keep you posted.

Ángel

Issue #92, opened 2020-12-04
[PW-3621] Custom fields for e.g. transaction ID not stored?


AndreasA:
**Describe the bug**
It appears the custom fields like PSP transaction ID are not stored on the order transaction.
Not sure, if that is by design.

**Expected behavior**
Store custom fields on order transaction.
Also way not store the current PSP reference ID too. why only the original?

**Additional context**

You execute: `$transaction->getOrderTransaction()->setCustomFields($customFields);`
However, the Shopware DAL is not an ORM and therefore this does not store the fields at all. You just set it in the order transaction object.

In order to store custom data on the order transaction you need to do something like this:
```php
        $context->scope(
            Context::SYSTEM_SCOPE,
            function (Context $context) use ($transaction, $customFields): void {
                   $this->orderTransactionRepository->update(
            [
                [
                    'id' => $transaction->getId(),
                    // Shopware uses JSON SQL functions to update only provided fields.
                    'customFields' => $cusotmFields,
                ],
            ],
            $context,
        );
            },
        );
```

acampos1916:
Thanks for opening this issue @AndreasA,

I've created an internal ticket to work on it. Your snippet will be useful!

Best,
Ángel

AndreasA:
The system scope is most likely necessary as I think you cannot update orders directly otherwise but you can try without it too.

PR #122, opened 2021-01-13
[PW-3768] replace usage of sales-channel-api with store-api

peterojo:
## Summary

Sales Channel API is being deprecated in S6.4 https://docs.shopware.com/en/shopware-platform-dev-en/sales-channel-api/sales-channel-context-api
Reimplement all the methods in SalesChannelApiController in a new custom StoreApiController and under the store-api scope.
Update route references in PaymentSubscriber and generate checkoutOrderUrl using store-api
## Tested scenarios

Checkout flow

**Fixed issue**:

PR #129, opened 2021-01-29
[PW-3922] Fix problem with completing a failed payment

peterojo:
## Summary

From the issue created in GitHub (https://github.com/Adyen/adyen-shopware6/issues/127)
## Tested scenarios

Pay with failing card details, then pay with one of redirect payment methods.
Pay with failing  redirect payment method, then pay with another redirect PM
Pay with failing card, the pay with valid card details

**Fixed issue**: #127 #92

### DIFF
--- a/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
@@ -92,7 +92,7 @@ export default class ConfirmOrderPlugin extends Plugin {
             // TODO error handling
             return;
         }
-        window.orderId = !!order.data ? order.data.id : order.id;
+        window.orderId = order.id;
         const finishUrl = new URL(
             location.origin + adyenCheckoutOptions.paymentFinishUrl);
         finishUrl.searchParams.set('orderId', order.id);
--- a/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
+++ b/src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
@@ -83,6 +83,8 @@ export default class ConfirmOrderPlugin extends Plugin {
     }
 
     afterCreateOrder(response) {
+        //response.subscribe(res =>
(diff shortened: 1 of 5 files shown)


cyattilakiss (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -92,7 +92,7 @@ export default class ConfirmOrderPlugin extends Plugin {
             // TODO error handling
             return;
         }
-        window.orderId = !!order.data ? order.data.id : order.id;
+        window.orderId = order.id;
```
acampos1916, line 95: This wasn't working for me a couple of weeks back and that's why I changed it here https://github.com/Adyen/adyen-shopware6/pull/125/files

But now with this PR I cannot reproduce the issue, so maybe it was my setup. Was this causing problems for you and that's why you undid it @peterojo?

acampos1916 (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: Oh you're right @acampos1916,  I thought I was the one who I added this when I [switched SalesChannel API for store API](https://github.com/Adyen/adyen-shopware6/pull/122/files#diff-ebd93096f9268cf9ecee4a9e5b146d5968c0fdcb0234dfd4bd284b8fcfd05403L95), I think it's okay to remove this check. The response from the `store-api` endpoint removes the `data` property and just returns the order in the root. 
So it's now `{id:xxxx,orderNumber:"yyyy",...}` instead of `{data:{id:xxxx,orderNumber:"yyyy"...}}`
So maybe you didn't yet have that change when you tested it?

acampos1916: Exactly! Thanks @peterojo

PR #131, opened 2021-02-15
[PW-3983] small fix, validate details in resultHandler

peterojo:
## Summary

[PW-3983] is solved by setting the `checkoutThreeDS1AuthorisationBehavior` ADP to `Auto` in the Customer Area.
This PR only introduces a minor change to check that the payment details exist before attempting to send the request to the [/payment/details API](https://docs.adyen.com/api-explorer/#/CheckoutService/v66/post/payments/details)
## Tested scenarios

3DS1 payment on Chrome v80+

### DIFF
--- a/src/Handlers/PaymentResponseHandler.php
+++ b/src/Handlers/PaymentResponseHandler.php
@@ -198,7 +198,7 @@ class PaymentResponseHandler
             return;
         }
 
-        // Get already stored transaction custom fileds
+        // Get already stored transaction custom fields
         $storedTransactionCustomFields = $transaction->getOrderTransaction()->getCustomFields() ?: [];
 
         // Store action, additionalData and originalPspReference in the transaction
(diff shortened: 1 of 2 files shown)


cyattilakiss (approved):

src/Handlers/PaymentResponseHandler.php
```diff
@@ -217,7 +217,7 @@ public function handleShopwareApis(
         }
 
         // Only store additional data for the transaction if this is the first additional data
-        $additionalData = $this->paymentResponseHandlerResult->getAction();
+        $additionalData = $this->paymentResponseHandlerResult->getAdditionalData();
```
acampos1916, line 220: Why were we getting the action in the first place?

peterojo: it was a typo, we get action a few lines above

cyattilakiss: pretty sure by mistake. The variable name also says additionalData :/

acampos1916 (approved):

PR #132, opened 2021-02-16
[PW-3652] bug fix in payment methods filter

peterojo:
## Summary

Fixed bug in payment methods filter for checkout page, that allowed PMs not returned by the `/paymentMethods` call to still be displayed in checkout
## Tested scenarios

iDeal does not show up for Belgium address
OneClick PMs do show up if they exist

### DIFF
--- a/src/Subscriber/PaymentSubscriber.php
+++ b/src/Subscriber/PaymentSubscriber.php
@@ -328,14 +328,13 @@ class PaymentSubscriber implements EventSubscriberInterface
         PaymentMethodCollection $originalPaymentMethods,
         SalesChannelContext $salesChannelContext,
         string $adyenPluginId
-    ): PaymentMethodCollection
-    {
+    ): PaymentMethodCollection {
         // Get Adyen /paymentMethods response
         $adyenPaymentMethods = $this->paymentMethodsService->getPaymentMethods($salesChannelContext);
 
         // If the /paymentMethods response returns empty, remove all Adyen payment methods from the list and return
-        if(empty($adyenPaymentMethods['paymentMethods'])) {
-            return $originalPaymentMethods->filter(function(PaymentMethodEntity $item) use ($adyenPluginId) {
+        if (empty($adyenPaymentMethods['paymentMethods'])) {
+            return $originalPaymentMethods->filter(function (PaymentMethodEntity $item) use ($adyenPluginId) {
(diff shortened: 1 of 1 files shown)


src/Subscriber/PaymentSubscriber.php
```diff
@@ -315,45 +318,50 @@ public function onCheckoutConfirmLoaded(PageLoadedEvent $event)
...
+                        $originalPaymentMethods->remove($paymentMethodEntity->getId());
                     }
-                );
-
-                $shopwareMethodIsOneClick = $pmCode == OneClickPaymentMethodHandler::getPaymentMethodCode();
-                $oneClickInResponse = !empty(
-                    $adyenPaymentMethods[OneClickPaymentMethodHandler::getPaymentMethodCode()]
-                );
+                } else {
+                    $paymentMethodFoundInResponse = array_filter(
```
cyattilakiss, line 354: can you pls add some comment here? It's not that easy to read code

cyattilakiss (dismissed):

src/Subscriber/PaymentSubscriber.php
```diff
@@ -315,45 +318,50 @@ public function onCheckoutConfirmLoaded(PageLoadedEvent $event)
...
+                        $adyenPaymentMethods['paymentMethods'],
+                        function ($value) use ($pmCode) {
+                            return $value['type'] == $pmCode;
+                        }
+                    );
 
-                //Remove the PM if it isn't in the paymentMethods response
-                //and if it is OneClick while not present in the response
-                if (empty($methodFoundInResponse) && ($shopwareMethodIsOneClick && !$oneClickInResponse)) {
```
peterojo, line 355: This logic only removed onclick when oneclick is not in the response, but ignored all other PMs where `$methodFoundInResponse` was empty

acampos1916 (changes requested):
Could you also add OneClick PMs to the tested scenarios? Basically checking if OneClick PMs do show up

peterojo:
> Could you also add OneClick PMs to the tested scenarios? Basically checking if OneClick PMs do show up

@acampos1916 yes this is also tested and they do show up. I've updated the PR description.

cyattilakiss (approved):

acampos1916 (approved):

PR #134, opened 2021-02-17
[PW-3736] fix broken test configuration link in shopware v6.3.5.1

peterojo:
## Summary

The Test configuration button in the admin panel is not shown correctly
Also, clicking the button returns an error.
This plugin changes the way we fetch the plugin config from the parent components,
allows for backward compatibility
## Tested scenarios

**Fixed issue**:


src/Resources/app/administration/src/component/adyen-config-check-button/adyen-config-check-button.html.twig
```diff
@@ -1,8 +1,9 @@
...
+<template>
     <sw-button-process
+            :disabled="isLoading"
             :isLoading="isLoading"
             :processSuccess="isSaveSuccessful"
             @process-finish="saveFinish"
             @click="check"
-    >{{ label }}</sw-button-process>
-</div>
+    >Test Configuration</sw-button-process>
```
peterojo, line 8: From v6.3.4.1 the {{label}} is no longer being displayed on the button in the Admin, moving the text from the config.xml to the twig template works across all the versions.

acampos1916: Nice. Can it also be replaced with a snippet so it can be translated? https://github.com/Adyen/adyen-shopware6/tree/develop/src/Resources/snippet

acampos1916: Now that I think about it the whole form is not translated anyway. So maybe we do the whole translation as a separated low prio ticket.

acampos1916 (approved):

src/Resources/app/administration/src/component/adyen-config-check-button/adyen-config-check-button.html.twig
peterojo: yeah that was my thinking as well

cyattilakiss (approved):

PR #139, opened 2021-02-23
[PW-3938] Using TaxDetector to subtract tax from order lines

acampos1916:
## Summary
If the customer belongs to a "Tax display = Net" group the order line's price is already tax exclusive. In this PR a check with TaxDetector is used to include or exclude the tax from the lines in the payments request.

## Tested scenarios
"Tax display = Net" group, the tax is not subtracted from the order lines.
"Tax display = Gross" group, the tax is subtracted from the order lines.

**Fixed issue**: PW-3938

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -43,6 +43,7 @@ use Adyen\Shopware\Storefront\Controller\RedirectResultController;
 use Adyen\Util\Currency;
 use Exception;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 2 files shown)


peterojo (approved):

cyattilakiss (approved):

AndreasA:
@acampos1916 @peterojo Shouldn't you use the `taxStatus` on the order to determine net prices instead of the tax detector because in theory that value could have been modified differently too though unlikely.

acampos1916:
@AndreasA thanks for the heads up! Indeed that would be easier.

PR #145, opened 2021-03-04
PW-4180 allow saving state data without paymentMethodId

### LINKED ISSUE #135: [PW-4180] Multi-Step / Custom checkouts (e.g. with headless) - Various issues
**Describe the bug**
Not sure, if I have found them all yet but in one project, the checkout is setup like this:
- Step 1:  Enter shipping method and addresses
- Step 2: Select payment method
- Technicall still step 2 but only after the payment method has been selected: Enter payment details
- Step 3: Summary plus entering additional data (one is required)

This results in:
- The adyen payment data is deleted as soon as one enters data in step 3 as you clear the context on all context changes which can also just be a currency change: https://github.com/Adyen/adyen-shopware6/blob/1.5.0/src/Subscriber/PaymentSubscriber.php#L370
- Saving the adyen data requires the payment method ID to be sent with it. However, as mentioned above, this can be done in a separate request and then it does not make sense to sent it again: https://github.com/Adyen/adyen-shopware6/blob/1.5.0/src/Subscriber/PaymentSubscriber.php#L341

Possible solutions:
- Use payment method ID from context, if the parameter `paymentMethodId` has not been provided as that is the current valid payment method ID then as it didn't change.
- Clear the adyen payment only for some context changes, e.g. if the payment method changes. Probably the only one it should be cleared actually.
- Add a scheduled task (like you already did for other stuff) that periodically checks if there are old/expired entries and remove them to avoid old payment data from incomplete checkouts filling up the database table. Maybe in combination with the context lifetime parameter: https://github.com/shopware/platform/blob/v6.3.5.1/src/Core/System/SalesChannel/Context/SalesChannelContextPersister.php#L35

peterojo:
## Summary

This PR allows sending of payment method and state data in separate requests, e.g for a multi-step payment form.
If the paymentMethodId is sent, then the current state data is cleared
If the state data is sent and the selected payment method is from Adyen, then it is stored
## Tested scenarios

Select PM without state data => state data is cleared if any
Select PM with state data => state data is saved
Select the same PM and update payment details => state data is replaced
Send only state data => state data is replaced (if available) or stored.

**Fixed issue**:  Closes #135


src/Subscriber/PaymentSubscriber.php
```diff
@@ -199,9 +208,32 @@ public static function getSubscribedEvents(): array
...
     {
-        // Save (or clear) state.data if payment method is selected/updated
+        // Clear state.data if payment method is updated
         if ($event->getRequestDataBag()->has('paymentMethodId')) {
-            $this->saveStateData($event);
+            $this->removeCurrentStateData($event);
+        }
+
+        // Save state data, only if Adyen payment method is selected
+        if ($event->getRequestDataBag()->get('adyenStateData')) {
```
AndreasA, line 217: See: https://github.com/Adyen/adyen-shopware6/issues/135#issuecomment-790900091

src/Subscriber/PaymentSubscriber.php
```diff
@@ -199,9 +208,32 @@ public static function getSubscribedEvents(): array
...
+            // Use payment method selected in the same request if available, otherwise get payment method from context
+            $paymentMethodId = $event->getRequestDataBag()->get('paymentMethodId')
+                ?? $event->getSalesChannelContext()->getPaymentMethod()->getId();
+            if (!$paymentMethodId) {
+                $this->logger->error('A payment method must be selected before saving payment state data.');
+                $this->session->getFlashBag()
```
AndreasA, line 223: btw. not really relevant but could you maybe use the validation scheme like the ContextSwitchroute does because upon using the store API, the flash messages in the session are completely irrelevant 😄

src/Subscriber/PaymentSubscriber.php
```diff
@@ -199,9 +216,35 @@ public static function getSubscribedEvents(): array
...
+            $this->removeCurrentStateData($event);
+        }
+
+        // Save state data, only if Adyen payment method is selected
+        if ($event->getRequestDataBag()->get('adyenStateData')) {
+            if (!$event->getRequestDataBag()->has('paymentMethodId')) {
+                // Validate that payment method has previously been saved.
+                $definition = new DataValidationDefinition('context_switch');
+                $definition->add(
+                    'paymentMethod',
```
AndreasA, line 230: shouldn't this be `paymentMethodId`? However, see above.

src/Subscriber/PaymentSubscriber.php
```diff
@@ -215,15 +223,18 @@ public function onContextTokenUpdate(SalesChannelContextSwitchEvent $event)
 
         // Save state data, only if Adyen payment method is selected
         if ($event->getRequestDataBag()->get('adyenStateData')) {
+            if (!$event->getRequestDataBag()->has('paymentMethodId')) {
```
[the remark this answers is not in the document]
AndreasA: maybe add an error similar to this to the `else` part of the condition `if ($paymentMethod->getPluginId() === $this->adyenPluginId) {` ? because then adyen payment data was provided for a non adyen payment method.

src/Subscriber/PaymentSubscriber.php
```diff
@@ -199,9 +205,29 @@ public static function getSubscribedEvents(): array
...
+            /** @var PaymentMethodEntity $paymentMethod */
+            $paymentMethod = $this->paymentMethodRepository->search(
+                (new Criteria())
+                    ->addFilter(new EqualsFilter('id', $paymentMethodId)),
+                $event->getContext()
+            )->first();
+            if ($paymentMethod->getPluginId() === $this->adyenPluginId) {
+                $this->saveStateData($event, $paymentMethod);
+            } else {
+                $this->logger->error('Please select an Adyen payment method to save payment state data.');
```
acampos1916, line 227: Since this is a log message maybe this is better?
```suggestion
                $this->logger->error('No Adyen payment method selected, skipping state data save.');
```

msilvagarcia (approved):

src/Subscriber/PaymentSubscriber.php
peterojo: You're right

msilvagarcia (dismissed):

src/Subscriber/PaymentSubscriber.php
```diff
@@ -380,58 +401,49 @@ private function trans(string $snippet, array $parameters = []): string
...
+        //Only store the state data if it matches the selected PM
+        if ($stateDataIsStoredPM == $selectedPaymentMethodIsStoredPM) {
+            try {
+                $this->paymentStateDataService->insertPaymentStateData(
+                    $event->getSalesChannelContext()->getToken(),
+                    $event->getRequestDataBag()->get('adyenStateData'),
+                    $event->getRequestDataBag()->get('adyenOrigin')
                 );
+            } catch (AdyenException $exception) {
+                $this->session->getFlashBag()
```
AndreasA, line 433: @peterojo Using the session flash bag here means that the store API will not be notified about it. maybe you can use the validation for it like the switch route itself does?

If not, I guess it isn't that big an issue if the store api does not receive feedback.

src/Subscriber/PaymentSubscriber.php
```diff
@@ -380,58 +401,49 @@ private function trans(string $snippet, array $parameters = []): string
...
-
-            //Payment method selected from the Shopware methods form
-            $selectedPaymentMethod = $this->paymentMethodRepository->search(
-                (new Criteria())
-                    ->addFilter(new EqualsFilter('id', $event->getRequestDataBag()->get('paymentMethodId'))),
-                Context::createDefaultContext()
-            )->first();
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            $this->logger->error('Payment state data is an invalid JSON: ' . json_last_error_msg());
+            $this->session->getFlashBag()
```
AndreasA, line 414: @peterojo Using the session flash bag here means that the store API will not be notified about it. maybe you can use the validation for it like the switch route itself does?

If not, I guess it isn't that big an issue if the store api does not receive feedback.

src/Subscriber/PaymentSubscriber.php
```diff
@@ -180,6 +182,10 @@ public function __construct(
         $this->cartCalculator = $cartCalculator;
         $this->currency = $currency;
         $this->logger = $logger;
+        $this->adyenPluginId = $this->pluginIdProvider->getPluginIdByBaseClass(
```
AndreasA, line 185: @peterojo are you sure this should be done in the constructor because due to dependency injection etc. this might happen without it being needed. Maybe add a getAdyenPluginId that retrieves the plugin ID the first time and stores it as variable in the class once it has been resolved?

src/Subscriber/PaymentSubscriber.php
```diff
@@ -380,58 +401,49 @@ private function trans(string $snippet, array $parameters = []): string
...
-                    $this->session->getFlashBag()
-                        ->add('danger', $this->trans('adyen.paymentMethodSelectionError'));
-                    return;
-                }
-            } else {
-                //PM selected and state.data don't match, clear previous state.data
-                $this->paymentStateDataService->deletePaymentStateDataFromContextToken(
-                    $event->getSalesChannelContext()->getToken()
+        //Only store the state data if it matches the selected PM
+        if ($stateDataIsStoredPM == $selectedPaymentMethodIsStoredPM) {
```
AndreasA, line 425: not really relevant here but just wondering any reason why you only use `==` and not `===`? the second one ensures the type matches too. I am mentioning this because e.g. `'1' == 1` results in `true` but `'1' === 1` results in `false`. The `===` is also a bit faster as it does not need to convert types.

AndreasA, line 425: @peterojo not really relevant here but just wondering any reason why you only use `==` and not `===`? the second one ensures the type matches too. I am mentioning this because e.g. `'1' == 1` results in `true` but `'1' === 1` results in `false`. The `===` is also a bit faster as it does not need to convert types.

src/Subscriber/PaymentSubscriber.php
peterojo: @AndreasA for now there are only 2 public functions triggered by separate events and they are will both need to instantiate this property when they are called, but I think your suggestion is very good and we can add it as a future improvement if we continue to expand this class

src/Subscriber/PaymentSubscriber.php
peterojo: Indeed, at the moment only we have the logs that the storeAPI client can use for debugging, but I am investigating what it takes to add store API responses for all our error messages throughout the plugin

src/Subscriber/PaymentSubscriber.php
peterojo: https://github.com/Adyen/adyen-shopware6/pull/145/files#r593041239

src/Subscriber/PaymentSubscriber.php
peterojo, line 425:
```suggestion
        if ($stateDataIsStoredPM === $selectedPaymentMethodIsStoredPM) {
```

msilvagarcia (approved):

acampos1916 (approved):

src/Subscriber/PaymentSubscriber.php
AndreasA: @peterojo Ah. Thanks. as mentioned not that big an issue anyway.

src/Subscriber/PaymentSubscriber.php
AndreasA: @peterojo Sure. Maybe it would actually make sense to create a computed/runtime property on the payment method itself that provides adyen information like: if it is an adyen payment method and/or what type. Could be useful upon retrieving the payment method using the store api too.

I mostly mentioned it because Symfony does a lot of stuff upon determining its dependency injection build process, e.g. service inlining if a service is only used once. Shouldn't be an issue but just to bear in mind 😄

PR #151, opened 2021-03-25
PW-4183 use event subscriber to instead of overriding edit payment method controller

### LINKED ISSUE #138: [PW-4183] Use EventSubscriber instead of replacing Storefront change payment method route
**Is your feature request related to a problem? Please describe.**
Currently you replace the Storefront change payment method route to ensure you can set the adyen data there.

However, this is can be problematic, if e.g. other plugins would do this too, e.g. multiple payment plugins.

**Describe the solution you'd like**
These are not the only solutions but some possibilities:
- Create an EventSubscriber that listens to `KernelEvents::REQUEST`:  https://symfony.com/doc/current/reference/events.html#kernel-request, checks `$request->attributes->get('_route')` to determine it is `frontend.account.edit-order.change-payment-method`. If it is, check if the adyen payment data is provided and then call your custom logic and `setResponse`  to the corresponding `RedirectResponse`.
- Alternatively, might be simpler or necessary as Shopware determines various contexts with those events which you might need: Use a controller event (ensure correct priority according to KernelListenerPriorities from Shopware) https://symfony.com/doc/current/reference/events.html#kernel-controller and ìf the correct controller/route is used and the adyen payment data, use `setController` to have it use your controller instead. See also https://symfony.com/doc/current/components/http_kernel.html#component-http-kernel-kernel-controller and https://symfony.com/doc/current/event_dispatcher/before_after_filters.html#creating-an-event-subscriber

peterojo:
## Summary

Fix for issue #138 
## Tested scenarios

**Fixed issue**:  resolves #138


cyattilakiss (approved):

acampos1916 (approved):

src/Subscriber/PaymentSubscriber.php
```diff
@@ -337,6 +360,35 @@ public function onCheckoutConfirmLoaded(PageLoadedEvent $event)
...
+            && $request->request->has('adyenStateData')) {
+            $this->contextSwitchRoute->switchContext(
+                new RequestDataBag(
+                    [
+                        SalesChannelContextService::PAYMENT_METHOD_ID => $request->get('paymentMethodId'),
+                        'adyenStateData' => $request->request->get('adyenStateData'),
+                        'adyenOrigin' => $request->request->get('adyenOrigin')
+                    ]
+                ),
+                $this->salesChannelContextFactory->create(
```
AndreasA, line 376: @peterojo Just a suggestion, in case you use `\Symfony\Component\HttpKernel\KernelEvents::CONTROLLER` event, with the priority `\Shopware\Core\Framework\Routing\KernelListenerPriorities::KERNEL_CONTROLLER_EVENT_CONTEXT_RESOLVE_POST` you could retrieve the sales channel context object from the request
` $request->attributes->get\Shopware\Core\PlatformRequest::ATTRIBUTE_SALES_CHANNEL_CONTEXT_OBJECT`);`

the problem, of course, then is that you cannot just execute `setResponse` but would need to call `setController` and call a custom controller.  Sadly, shopware does everything regarding context resolving etc. in the controller event which I think should actually be in the request event.

P.S. I just thought about a solution that migh be better. You could use service decoration (https://symfony.com/doc/current/service_container/service_decoration.html) for the ContextSwitchRoute by creating a custom class that extends `\Shopware\Core\System\SalesChannel\SalesChannel\AbstractContextSwitchRoute::switchContext`, sets the `RouteScope` at the top of the class but no `Route` definition as that is not necessary. 

Now you could check if you are in the route 'frontend.account.edit-order.change-payment-method' by retrieving the current request from the request stack service (https://symfony.com/doc/current/service_container/request.html) and if so adjust the request databag to include your parameters.
This would allow you to keep the remaining funcionality Shopware default behaviour including the redirect etc.

P.S.2: if you do not remove the adyen state data, if the payment method ID stays the same, you would only need to execute the switch route here and the rest would work with Shopware default too. But to be honest, I like the service decoration approach the most, as it uses just default behaviour.

peterojo: @AndreasA I tried the approach of decorating `ContextSwitchRoute`, but I get the error `Argument 3 passed to Shopware\Storefront\Controller\AccountOrderController::__construct() must be an instance of Shopware\Core\System\SalesChannel\SalesChannel\ContextSwitchRoute, instance of Adyen\Shopware\Service\ContextSwitchRoute given`, because I would need to also decorate or override the `AccountOrderController` and that's what this ticket is changing to begin with. For now the current approach seems to me to be the most clear cut, even though I had to inject the salesChannelFactory. I can refactor later if there's a better option.

AndreasA: 1.)
Ok. No it is fine. It seems that the issue is actually in the PaymentController where Shopware uses the wrong typehinting. It should be the AbstractContextSwitchRoute.

The only thing that might be changed now is to use the `Controller` event that way you could access the context.

If you add that setting the same payment method ID does not trigger the payment method removal you could just set the adyen state and let everything else be done by the default controller.

2.)
Or alternatively, use the controller event and if `adyenStateData` is available in the request, adjust the controller to call a custom adyen controller which does the context switching accordingly.

but as it is now is fine too.

AndreasA: @peterojo 

1.)
Ok. No it is fine. It seems that the issue is actually in the PaymentController where Shopware uses the wrong typehinting. It should be the AbstractContextSwitchRoute.

The only thing that might be changed now is to use the `Controller` event that way you could access the context.

If you add that setting the same payment method ID does not trigger the payment method removal you could just set the adyen state and let everything else be done by the default controller.

2.)
Or alternatively, use the controller event and if `adyenStateData` is available in the request, adjust the controller to call a custom adyen controller which does the context switching accordingly.

but as it is now is fine too.

Though personally I think point 1 would be the least intrusive method to do this as the default controller is still called aferwards, you just do stuff beforehand.

PR #153, opened 2021-03-29
[PW-4117] exclude discount from line items in open invoice request

peterojo:
## Summary

When the shopping cart contains items with different tax rates and a discount is applied, then the `getCalculatedTaxes()` value of the discount contains multiple entities, with the current approach of picking the `->first()` taxRate, we get a wrong result and the API returns an error similar to `Acquirer Error (BAD_VALUE : order_lines[6].total_tax_amount)`.
API allows sending the line items without the discount https://docs.adyen.com/payment-methods/klarna/invoice-lines#no-discount-line, so this is the simplest way to solve the problem
## Tested scenarios

Send payment request with multiple line items of different tax rates and apply discount

**Fixed issue**:

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -549,15 +549,14 @@ abstract class AbstractPaymentMethodHandler
                 //Getting line price
                 $price = $orderLine->getPrice();
 
-                //Getting order line information differently if it's a promotion or product
+                // Skip promotion line items.
                 if (empty($orderLine->getProductId()) && $orderLine->getType() === self::PROMOTION) {
-                    $productName = $orderLine->getDescription();
-                    $productNumber = $orderLine->getPayload()['promotionId'];
-                } else {
-                    $product = $this->getProduct($orderLine->getProductId(), $salesChannelContext->getContext());
-                    $productName = $product->getName();
-                    $productNumber = $product->getProductNumber();
+                    continue;
                 }
+
+                $product = $this->getProduct($orderLine->getProductId(), $salesChannelContext->getContext());
+                $productName = $product->getName();
+                $productNumber = $product->getProductNumber();
 
                 //Getting line tax amount and rate
                 $lineTax = $price->getCalculatedTaxes()->getAmount() / $orderLine->getQuantity();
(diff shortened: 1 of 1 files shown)


src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -568,23 +567,21 @@ private function preparePaymentsRequest(
...
 
+                $currency = $this->getCurrency(
+                    $transaction->getOrder()->getCurrencyId(),
+                    $salesChannelContext->getContext()
+                );
                 //Building open invoice line
                 $lineItems[] = $this->openInvoiceBuilder->buildOpenInvoiceLineItem(
                     $productName,
                     $this->currency->sanitize(
                         $price->getUnitPrice() -
```
cyattilakiss, line 578: does the unit price already include the discount? 
I can imagine in case it's a percentage discount then yes but what if it's like 5 eur discount from the order total, is it still handled well in this case?

peterojo: No the price for each line item will not include the discount, normally the discount will be added as one of the line items with a negative value, e.g `"amountExcludingTax":"-120"`, and the total added up will be equal to the amount in `amount.value` of the request. But in this case, the discount is not included, and the total amount (adding up the line items) will be greater that the figure in `amount.value`.
```json
...
"amount": {
    "currency": "EUR",
    "value": 3000
  },
"lineItems": [
    {
      "id": "feac1274086a4531a9755492eb135102",
      "description": "Rustic Steel Aqua Radiant",
      "amountExcludingTax": 50467,
      "taxAmount": 3533,
      "taxPercentage": 700,
      "quantity": 1,
      "taxCategory": ""
    },
    {
      "id": "fe3b463d1aac4e1dbd4007b7dc47bb88",
      "description": "Aerodynamic Bronze iPlug",
      "amountExcludingTax": 88700,
      "taxAmount": 0,
      "taxPercentage": 0,
      "quantity": 1,
      "taxCategory": ""
    },
    {
      "id": "f2f6272cfb8d46d485020aa3cfca6712",
      "description": "Aerodynamic Cotton breezz",
      "amountExcludingTax": 56555,
      "taxAmount": 10745,
      "taxPercentage": 1900,
      "quantity": 1,
      "taxCategory": ""
    }
  ],
```
But the API will figure it out and add the discount line in the request to klarna

acampos1916 (approved):

src/Handlers/AbstractPaymentMethodHandler.php
cyattilakiss: let's create a follow up ticket for this because indeed the API will adjust the value but it's not the best approach to rely on only the API :(

cyattilakiss (approved):

PR #161, opened 2021-04-15
[PW-4474] add refund notification processor

### LINKED ISSUE #157: [PW-4474] Payment Refund
**Is your feature request related to a problem? Please describe.**
Whenever we refund a user from Adyen admin panel and the status doesn't change on Shopware backend.

**Describe the solution you'd like**
- We want to be able to make an order refund from Shopware backend by changing the order status.
- If we refunded a payment from Adyen admin panel the order status is updated at Shopware backend.

peterojo:
## Summary

Create REFUND notification processor.
Transition order transactions to refunded/refunded_partially when notification is received and handle illegal state transitions
## Tested scenarios

open -> refunded (illegal: moves to paid and then to refunded) ✅
refunded -> refunded (illegal: skips) ❌
refunded -> [other] (illegal: skips) ❌
paid -> refunded ✅
cancelled -> refunded ✅
refunded_partially -> refunded ✅

**Fixed issue**:  Resolves #157

### DIFF
--- a/src/NotificationProcessor/RefundNotificationProcessor.php
+++ b/src/NotificationProcessor/RefundNotificationProcessor.php
@@ -49,6 +49,12 @@ class RefundNotificationProcessor extends NotificationProcessor implements Notif
                 $orderTransaction->getAmount()->getTotalPrice(),
                 $this->getOrder()->getCurrency()->getIsoCode()
             );
+
+            if ($refundedAmount > $transactionAmount) {
+                $this->logger->warning('The refunded amount is greater than the transaction amount.', $logContext);
+                return;
+            }
+
             $partial = $refundedAmount < $transactionAmount;
             $newState = $partial
                 ? OrderTransactionStates::STATE_PARTIALLY_REFUNDED
--- a/src/NotificationProcessor/RefundNotificationProcessor.php
+++ b/src/NotificationProcessor/RefundNotificationProcessor.php
@@ -55,8 +55,7 @@ class RefundNotificationProcessor extends NotificationProcessor implements Notif
                 return;
             }
 
-            $partial = $refundedAmount < $transactionAmount;
-            $newState = $partial
+            $newState = $refundedAmount < $transactionAmount
                 ? OrderTransactionStates::STATE_PARTIALLY_REFUNDED
                 : OrderTransactionStates::STATE_REFUNDED;
             // Transactions in refunded state cannot be transitioned to any other state,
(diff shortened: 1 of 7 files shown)


msilvagarcia (commented):
- The `currencyUtil` property in the `NotificationProcessor` class is not actually used.
- Why change the return value of `NotificationProcessorFactory#create` from `NotificationProcessorInterface` to `NotificationProcessor`?
- Should we worry about refund values?
-

msilvagarcia (changes requested):
https://github.com/Adyen/adyen-shopware6/pull/161#pullrequestreview-637412232

src/NotificationProcessor/RefundNotificationProcessor.php
```diff
@@ -0,0 +1,84 @@
...
+        $logContext = [
+            'orderId' => $this->getOrder()->getId(),
+            'orderNumber' => $this->getOrder()->getOrderNumber(),
+            'eventCode' => NotificationEventCodes::REFUND,
+            'originalState' => $state
+        ];
+
+        if ($this->getNotification()->isSuccess()) {
+            $refundedAmount = (int) $this->getNotification()->getAmountValue();
+            $transactionAmount = $this->currencyUtil->sanitize(
```
peterojo, line 48: @msilvagarcia currencyUtil is used here

peterojo:
@msilvagarcia 
> I love the refactor in the Order Repository. Great implementation :)
🍻
> * The `currencyUtil` property in the `NotificationProcessor` class is not actually used. Maybe that was something from another change, or something you planned to add and changed your mind? :)
it's used in the refund processor
> * Why change the return value of `NotificationProcessorFactory#create` from `NotificationProcessorInterface` to `NotificationProcessor`? Seems better to me to continue using the interface and update it accordingly if we need to add more methods, specially since we have multiple types of Notification Processors.
Good point, I should have left it alone, and implemented the interface on the abstract class `NotificationProcessor`, that's what was missing
> * Should we worry about refund values? How often does it occur that we receive a notification for a refund amount higher than the transaction amount?
The way I see it, we only care if the refund value is less because then it is a partial refund (Adyen just sends the same `REFUND` notification). If for some reason the refund amount is somehow greater than the transaction amount then it still counts as a full refund. What do you think @acampos1916

cyattilakiss:
> The way I see it, we only care if the refund value is less because then it is a partial refund (Adyen just sends the same REFUND notification). If for some reason the refund amount is somehow greater than the transaction amount then it still counts as a full refund. What do you think @acampos1916

@peterojo do we have a status like "needs attention" or anything like that? In case the amount is greater than the original one then it's definitely a potential issue. If possible I would flag it somehow while setting the status to refunded but indeed it should not happen.
If the amount is less then it's a partial refund indeed.

peterojo:
> @peterojo do we have a status like "needs attention" or anything like that? In case the amount is greater than the original one then it's definitely a potential issue. If possible I would flag it somehow while setting the status to refunded but indeed it should not happen.
> If the amount is less then it's a partial refund indeed.
Since we don't have such a status like 'needs attention', if the refunded amount is greater, then I will leave a warning log and *not* set the status to refunded. So that the merchant can be aware.

src/NotificationProcessor/RefundNotificationProcessor.php
```diff
@@ -0,0 +1,90 @@
...
+                $orderTransaction->getAmount()->getTotalPrice(),
+                $this->getOrder()->getCurrency()->getIsoCode()
+            );
+
+            if ($refundedAmount > $transactionAmount) {
```
cyattilakiss, line 53: Just realised maybe it's also a good idea to add this as an error in the notification record error_message field therefore it's a bit more visible for the merchants what is the problem and with which notification

Action for the merchant can be the following:
Check whether the refund is indeed for the order indicated in the merchantReference. If yes then manually move the order to refunded and check why the value difference

If it's not for that order then it's good that you don't refund it in shopware and then the merchant can reach out to support to figure out why the wrong connection

src/NotificationProcessor/RefundNotificationProcessor.php
```diff
@@ -0,0 +1,90 @@
...
+    private function doRefund(OrderTransactionEntity $orderTransaction, Context $context, bool $partial = false)
+    {
+        try {
+            if ($partial) {
+                $this->getTransactionStateHandler()->refundPartially($orderTransaction->getId(), $context);
+            } else {
+                $this->getTransactionStateHandler()->refund($orderTransaction->getId(), $context);
+            }
+        } catch (IllegalTransitionException $exception) {
+            // set to paid, and then try again
```
acampos1916, line 85: Maybe we could add a log here to let the merchant know that the transaction was moved to paid before being moved to refunded

src/NotificationProcessor/RefundNotificationProcessor.php
peterojo: that's a good point, in that case my solution would be to throw an exception here and it will be saved in the [handler](https://github.com/Adyen/adyen-shopware6/blob/df5929d343c71ae92adf751c190ea4d47c29d95f/src/ScheduledTask/ProcessNotificationsHandler.php#L138) as an error on the notification (also logged as an error).

cyattilakiss: perfect :)

cyattilakiss (approved):

acampos1916 (approved):

msilvagarcia (approved):

Issue #154, opened 2021-03-30
Extend shopware payment method entity to include adyen paymentmethods data.


peterojo:
**Headless integrations have to add workaround for getting /paymentMethods data**
With headless integrations such as PWAs it would be useful to extend each Adyen payment method entity to include the corresponding data from the `/paymentMethods` API response from adyen as well as fields like `useAdyen: true`, and `type:scheme`

**Proposed solution**
Create entity extension for PaymentMethodEntity and add new fields for `usesAdyen`, `type` as well as json field for the filtered `/paymentMethods` data for each PM.
Listen for requests to `/store-api/v*/payment-method` and update all adyen PMs with the latest data from the API
Find a way to include adyen info in the `/store-api/v*/payment-method` API response 

**Alternative solution**
Another solution would be to periodically update the payment method entities via a cronjob instead of listening for payment-method requests

peterojo:
@AndreasA I wonder if you have any ideas about implementing such a solution

AndreasA:
@peterojo One solution would be to use the event 'sales_channel.' . PaymentEvents::PAYMENT_METHOD_LOADED_EVENT

this would allow you to add data on the fly to the payment method response, if it uses the sales channel repository.

Maybe you also need to do so without sales channel and check if you are in the store-api etc. otherwise but I think it should be enough for the sales channel.

you will need an entity extension with a runtime computed field too.

AndreasA:
@peterojo Alternative, in this case it might be better as the payment method response is only required for this one API:

- Create a Response subscriber.
- Ensure it is executed before `\Shopware\Core\System\SalesChannel\Api\StoreApiResponseListener`
- Check if the response is `instanceof PaymentMethodRouteResponse`
- Maybe also check the `route` used is `store-api.payment.method` but with the `instanceof` check it is not necessary
- Add extension to payment methods in the `PaymentMethodRouteResponse`.

AndreasA:
@peterojo I hope I will be able to do create PRs this or next week. At least currently it looks good that I will get the time.

peterojo:
@AndreasA that would be great, we have also scheduled this ticket for this or next week so I will definitely allocate some time. Please go ahead and make a PR whenever you can and I will take a look.

AndreasA:
@peterojo One question: I should probably also limit the available payment methods here, correct?

[...]

PR #162, opened 2021-04-19
Add payment methods response and additional information to store api response

AndreasA:
## Summary
It adds adyen payment method information to the payment method store API response.

## Tested scenarios
Retrieving payment methods using Store-API.

**Fixed issue**:  #154

### DIFF
--- a/src/Core/Checkout/Order/Aggregate/OrderTransaction/OrderTransactionExtension.php
+++ b/src/Core/Checkout/Order/Aggregate/OrderTransaction/OrderTransactionExtension.php
@@ -25,7 +25,10 @@ namespace Adyen\Shopware\Core\Checkout\Order\Aggregate\OrderTransaction;
 
 use Adyen\Shopware\Entity\PaymentResponse\PaymentResponseEntityDefinition;
 use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionDefinition;
+use Shopware\Core\Framework\Api\Context\SalesChannelApiSource;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityExtension;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\ApiAware;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\ReadProtected;
 use Shopware\Core\Framework\DataAbstractionLayer\Field\OneToManyAssociationField;
 use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
 
@@ -41,12 +44,17 @@ class OrderTransactionExtension extends EntityExtension
 
     public function extendFields(FieldCollection $collection): void
     {
-        $collection->add(
-            new OneToManyAssociationField(
-                'adyenPaymentResponses',
-                PaymentResponseEntityDefinition::class,
-                'order_transaction_id'
-            )
+        $field = new OneToManyAssociationField(
+            'adyenPaymentResponses',
+            PaymentResponseEntityDefinition::class,
+            'order_transaction_id'
         );
+
+        // Ensure the data is not available via the Store API in older Shopware versions.
+        if (!class_exists(ApiAware::class) && class_exists(ReadProtected::class)) {
+            $field->addFlags(new ReadProtected(SalesChannelApiSource::class));
+        }
+
+        $collection->add($field);
     }
 }
--- a/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
+++ b/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
@@ -1,0 +1,51 @@
+<?php declare(strict_types=1);
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2020 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\Core\Checkout\Payment\PaymentMethod;
+
+use Shopware\Core\Checkout\Payment\PaymentMethodDefinition;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityExtension;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\ApiAware;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Computed;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Runtime;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\ObjectField;
+use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
+
+class PaymentMethodExtension extends EntityExtension
+{
+    public function extendFields(FieldCollection $collection): void
+    {
+        $collection->add(
+            (new ObjectField(
+                'adyen_data',
+                'adyenData'
+            ))->addFlags(new Runtime(), new Computed(), new ApiAware())
+        );
+    }
+
+    public function getDefinitionClass(): string
+    {
+        return PaymentMethodDefinition::class;
+    }
+}
--- a/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
+++ b/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
@@ -15,7 +15,7 @@
  *
  * Adyen Payment Module
  *
- * Copyright (c) 2020 Adyen B.V.
+ * Copyright (c) 2021 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  *
--- a/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
+++ b/src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
@@ -36,12 +36,17 @@ class PaymentMethodExtension extends EntityExtension
 {
     public function extendFields(FieldCollection $collection): void
     {
-        $collection->add(
-            (new ObjectField(
-                'adyen_data',
-                'adyenData'
-            ))->addFlags(new Runtime(), new Computed(), new ApiAware())
+        $field = new ObjectField(
+            'adyen_data',
+            'adyenData'
         );
+
+        $field->addFlags(new Runtime(), new Computed());
+        if (class_exists(ApiAware::class)) {
+            $field->addFlags(new ApiAware());
+        }
+
+        $collection->add($field);
     }
 
     public function getDefinitionClass(): string
--- a/src/Provider/AdyenPluginProvider.php
+++ b/src/Provider/AdyenPluginProvider.php
@@ -13,7 +13,7 @@
  *                               #############
  *                               ############
  *
- * Adyen plugin for Shopware 6
+ * Adyen Payment Module
  *
  * Copyright (c) 2021 Adyen B.V.
  * This file is open source and available under the MIT license.
--- a/src/Provider/AdyenPluginProvider.php
+++ b/src/Provider/AdyenPluginProvider.php
@@ -15,7 +15,7 @@
  *
  * Adyen plugin for Shopware 6
  *
- * Copyright (c) 2020 Adyen B.V.
+ * Copyright (c) 2021 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  *
--- a/src/Provider/AdyenPluginProvider.php
+++ b/src/Provider/AdyenPluginProvider.php
@@ -1,0 +1,66 @@
+<?php declare(strict_types=1);
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen plugin for Shopware 6
+ *
+ * Copyright (c) 2020 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ */
+
+namespace Adyen\Shopware\Provider;
+
+use Adyen\Shopware\AdyenPaymentShopware6;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\Plugin\Util\PluginIdProvider;
+use Symfony\Contracts\Service\ResetInterface;
+
+class AdyenPluginProvider implements ResetInterface
+{
+    /**
+     * @var string|null
+     */
+    protected $adyenPluginId = null;
+
+    /**
+     * @var PluginIdProvider
+     */
+    protected $pluginIdProvider;
+
+    public function __construct(PluginIdProvider $pluginIdProvider)
+    {
+        $this->pluginIdProvider = $pluginIdProvider;
+    }
+
+    public function getAdyenPluginId(): string
+    {
+        if (isset($this->adyenPluginId)) {
+            return $this->adyenPluginId;
+        }
+
+        $this->adyenPluginId = $this->pluginIdProvider->getPluginIdByBaseClass(
+            AdyenPaymentShopware6::class,
+            Context::createDefaultContext()
+        );
+
+        return $this->adyenPluginId;
+    }
+
+    public function reset()
+    {
+        $this->adyenPluginId = null;
+    }
+}
--- a/src/Resources/config/services/providers.xml
+++ b/src/Resources/config/services/providers.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+  ~                       ######
+  ~                       ######
+  ~ ############    ####( ######  #####. ######  ############   ############
+  ~ #############  #####( ######  #####. ######  #############  #############
+  ~        ######  #####( ######  #####. ######  #####  ######  #####  ######
+  ~ ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+  ~ ###### ######  #####( ######  #####. ######  #####          #####  ######
+  ~ #############  #############  #############  #############  #####  ######
+  ~  ############   ############  #############   ############  #####  ######
+  ~                                      ######
+  ~                               #############
+  ~                               ############
+  ~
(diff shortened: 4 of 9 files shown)


src/Struct/AdyenPaymentMethodDataStruct.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Adyen\Shopware\Struct;
+
+use Shopware\Core\Framework\Struct\Struct;
+
+class AdyenPaymentMethodDataStruct extends Struct
+{
+    /**
+     * @var bool
+     */
+    protected $isAdyen = false;
```
AndreasA, line 34: @peterojo I think we could actually remove this flag because I could just not add the extension at all for non adyen payment methods. However, this way one can be sure that it was not a but that there is not adyen data added to the payment method.

src/Subscriber/Response/PaymentMethodRouteResponseSubscriber.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->adyenPluginProvider = $adyenPluginProvider;
+        $this->paymentMethodsService = $paymentMethodsService;
+        $this->paymentMethodsFilterService = $paymentMethodsFilterService;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            // Ensure event is executed before "Shopware\Core\System\SalesChannel\Api\StoreApiResponseListener".
+            KernelEvents::RESPONSE => ['adjustPaymentMethodsResponse', 11000],
```
AndreasA, line 77: @peterojo I have for now done it with a response subscriber as I think this data is only relevant for that one Store API response route? 

If you want to have the same data on other responses that might include the payment method etc. I would need to use a different event or a service decoration. Both are possible but I think this should be fine?

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
```diff
@@ -0,0 +1,51 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
+
+class PaymentMethodExtension extends EntityExtension
+{
+    public function extendFields(FieldCollection $collection): void
+    {
+        $collection->add(
+            (new ObjectField(
+                'adyen_data',
+                'adyenData'
```
AndreasA, line 42: @peterojo Currently not strictly necessary but normally computed fields should be declared on an entity (if it is an entity) in its entity definition, so to be safe I added it.

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
```diff
@@ -0,0 +1,51 @@
...
+            ))->addFlags(new Runtime(), new Computed(), new ApiAware())
```
peterojo, line 43: is the `ApiAware` flag really necessary? I see from it's definition that it only has the following whitelisted:
```
AdminApiSource::class => '/api/',
SalesChannelApiSource::class => '/sales-channel-api/',
```
now that we no longer use sales-channel-api won't this be a problem?

src/Subscriber/Response/PaymentMethodRouteResponseSubscriber.php
```diff
@@ -0,0 +1,182 @@
...
+            $paymentMethodsResponse['paymentMethods'],
+            function ($value) use ($type) {
+                return ($value['type'] ?? null) == $type;
+            }
+        );
+
+        return $paymentMethodsResponse;
+    }
+
+    private function getPaymentMethodType(PaymentMethodEntity $method): ?string
```
peterojo, line 157: Very nice! 👍🏿

src/Provider/AdyenPluginProvider.php
```diff
@@ -0,0 +1,66 @@
+<?php declare(strict_types=1);
```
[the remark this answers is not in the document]
AndreasA: @peterojo It does not need to call the database, if it is accessed twice as I remember the plugin ID in the services memory which stays the same for the full request. probably not strictly necessary as Shopware also caches database queries etc. but it makes it a bit easier to retrieve the data in multiple places.

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
AndreasA: @peterojo actually since 6.3.5.1 it is required for this data to be shown in the store api. not sure where exactly they doe the mapping but it actually supports store api without any additional stuff.

However, if you want to support older SW versions then 6.3.5.1,  I will have to remove it but then it might be lost in 6.3.5 as they switched from blacklisting to whitelisting in 6.3.5.1

I could check if the class exists and only add the flag if it exists, then it should support new and old versions

src/Provider/AdyenPluginProvider.php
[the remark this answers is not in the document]
peterojo: fair enough

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
AndreasA: I now check for class existince to ensure it works with previous versions too. However, this flag is enough for the store api without any adjustments. tested in multiple projects now. Not sure why the base url for the store api is not specified in the flag though but it does work and this is how shopware does it internally now too in all their definitions.

src/Struct/AdyenPaymentMethodDataStruct.php
AndreasA: @peterojo removed it now.

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
peterojo: Great, we would prefer to still continue to support all versions from v6.3.3.1 and above for as long as possible

src/Provider/AdyenPluginProvider.php
```diff
@@ -0,0 +1,66 @@
...
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen plugin for Shopware 6
```
peterojo, line 16:
```suggestion
 * Adyen Payment Module
```
I just noticed that we still have `Adyen plugin for Shopware 6` in a few other places, but let's standardize on `Adyen Payment Module`

src/Resources/config/services/providers.xml
```diff
@@ -0,0 +1,34 @@
...
+  ~        ######  #####( ######  #####. ######  #####  ######  #####  ######
+  ~ ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+  ~ ###### ######  #####( ######  #####. ######  #####          #####  ######
+  ~ #############  #############  #############  #############  #####  ######
+  ~  ############   ############  #############   ############  #####  ######
+  ~                                      ######
+  ~                               #############
+  ~                               ############
+  ~
+  ~ Adyen plugin for Shopware 6
```
peterojo, line 16:
```suggestion
  ~ Adyen Payment Module
```

src/Service/PaymentMethodsFilterService.php
```diff
@@ -0,0 +1,96 @@
...
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen plugin for Shopware 6
```
peterojo, line 16:
```suggestion
 * Adyen Payment Module
```

src/Core/Checkout/Payment/PaymentMethod/PaymentMethodExtension.php
AndreasA: @peterojo Ok. i retested and without the ApiAware flag and 6.3.5.1 the data is missing. with it is shown. and the class check should ensure it still works fine for older version.

src/Core/Checkout/Order/Aggregate/OrderTransaction/OrderTransactionExtension.php
```diff
@@ -41,12 +44,17 @@ public function getDefinitionClass(): string
...
-                PaymentResponseEntityDefinition::class,
-                'order_transaction_id'
-            )
+        $field = new OneToManyAssociationField(
+            'adyenPaymentResponses',
+            PaymentResponseEntityDefinition::class,
+            'order_transaction_id'
         );
+
+        // Ensure the data is not available via the Store API in older Shopware versions.
```
AndreasA, line 53: @peterojo As you still support older versions, I added the readrpotect field here as otherwise the data could be retrieved using the store API in versions prior to 6.3.5.1

peterojo: @AndreasA this extension does not have the `ApiAware()` flag, so why is it necessary to block older versions?

AndreasA: @peterojo 
older versions of Shopware use inverted flags behavior. ReadProtected will be removed in future versions altogether.

- up to 6.3.5.0: It is necessary to specify which fields are not available via the Store api using the ReadProtected flag
- after 6.3.5.1 (probably wanted to do this in 6.3.5.0 already) and newer: It is necessary to mark all fields that should be available via Store API by using the ApiAware flag.

So to ensure older versions of Shopware cannot access the field using Store API, the ReadProtected flag has to be added.

peterojo (approved):

src/Service/PaymentMethodsFilterService.php
```diff
@@ -0,0 +1,96 @@
...
+namespace Adyen\Shopware\Service;
+
+use Adyen\Shopware\Handlers\OneClickPaymentMethodHandler;
+use Shopware\Core\Checkout\Payment\PaymentMethodCollection;
+use Shopware\Core\Checkout\Payment\PaymentMethodEntity;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+class PaymentMethodsFilterService
+{
+    private PaymentMethodsService $paymentMethodsService;
```
peterojo, line 33: Just caught this, typed properties are new in php 7.4, this would break older versions
```suggestion
    private $paymentMethodsService;
```

AndreasA: forgot to adjust to the pre PHP 7.4 here 😄 Thanks for catching it. we usually only work with 7.4 mostly now, so I really had to take care to adjust to pre 7.4 code patterns.

AndreasA: @peterojo forgot to adjust to the pre PHP 7.4 here 😄 Thanks for catching it. we usually only work with 7.4 mostly now, so I really had to take care to adjust to pre 7.4 code patterns.

should be fixed now.

peterojo (approved):

acampos1916 (approved):

peterojo:
Thanks @AndreasA 🎉

src/Subscriber/Response/PaymentMethodRouteResponseSubscriber.php
```diff
@@ -0,0 +1,182 @@
...
+            if (!empty($type)) {
+                $extension->setPaymentMethodsResponse($this->getPaymentMethodsResponseForType($context, $type));
+            }
+
+            $method->addExtension('adyenData', $extension);
+        }
+    }
+
+    private function getPaymentMethodsResponseForType(SalesChannelContext $context, string $type)
+    {
```
peterojo, line 141: @AndreasA I want to propose a revision for this method. Instead of returning the entire $paymentMethodsResponse, we could just return the data for a single payment method (ie `$paymentMethodsResponse['paymentMethods'][0]`) and return null if the payment method data is not found.
The current implementation returns the following:
```
...
"extensions": {
    ...
    "adyenData": {
        "type": "paypal",
        "paymentMethodsResponse": {
            "paymentMethods": {
                "1": {
                    "configuration": {
                        "merchantId": "MERCHANT_ACCOUNT",
                        "intent": "capture"
                    },
                    "name": "PayPal",
                    "type": "paypal"
                }
            }
        },
        "apiAlias": "adyen_shopware_struct_adyen_payment_method_data_struct"
    }
},
```
and I think we can instead change it to the following:
```
...
"extensions": {
    ...
    "adyenData": {
        "type": "paypal",
        "paymentMethodsResponse":  {
            "configuration": {
                "merchantId": "MERCHANT_ACCOUNT",
                "intent": "capture"
            },
            "name": "PayPal",
            "type": "paypal"
        },
        "apiAlias": "adyen_shopware_struct_adyen_payment_method_data_struct"
    }
},
```
What do you think? Will this have a big impact on your existing implementation?

AndreasA: @peterojo Don't we already limit the payment methods response to valid payment methods for that payment method?

And couldn't this still be an array as e.g. credit cards would have visa/mastercard etc.? or is this still just one?

AndreasA: @peterojo 
```
        $paymentMethodsResponse['paymentMethods'] = array_filter(
            $paymentMethodsResponse['paymentMethods'],
            function ($value) use ($type) {
                return ($value['type'] ?? null) == $type;
            }
        );
```

However, in case you are sure that there is always only ONE payment method valid per payment method then it could be refactored. 

However, doesn't adyen require the full payment methods response when creating the component? which is why with the current implementation one can just use
```
 paymentMethodsResponse: paymentMethod.extensions.adyenData.paymentMethodsResponse, 
```

However, in theory we can change this but not sure, if we then have all the necessary data for the web component call ?

AndreasA: @peterojo 

1.) Don't we already limit the payment methods response to valid payment methods for that payment method?

And couldn't this still be an array as e.g. credit cards would have visa/mastercard etc.? or is this still just one?

It should be done here:
```
        $paymentMethodsResponse['paymentMethods'] = array_filter(
            $paymentMethodsResponse['paymentMethods'],
            function ($value) use ($type) {
                return ($value['type'] ?? null) == $type;
            }
        );
```


2.) In case you are sure that there is always only ONE payment method valid per payment method then it could be refactored. 

3.) However, doesn't adyen require the full payment methods response when creating the component? which is why with the current implementation one can just use
```
 paymentMethodsResponse: paymentMethod.extensions.adyenData.paymentMethodsResponse, 
```

However, in theory we can change this but not sure, if we then have all the necessary data for the web component call ?

peterojo: @AndreasA 
1.) & 2.) The payment method `type` in the `/paymentMethods` response is unique to each PM, and it does not have to be an array. The response for credit cards would be:
```json
{
  "brands": [
    "visa",
    "mc"
  ],
  "details": [
    {
      "key": "encryptedCardNumber",
      "type": "cardToken"
    },
    {
      "key": "encryptedSecurityCode",
      "type": "cardToken"
    },
    {
      "key": "encryptedExpiryMonth",
      "type": "cardToken"
    },
    {
      "key": "encryptedExpiryYear",
      "type": "cardToken"
    },
    {
      "key": "holderName",
      "optional": true,
      "type": "text"
    }
  ],
  "name": "Credit Card",
  "type": "scheme"
}
```
3.)  However, you made a valid point about the [web component](https://docs.adyen.com/online-payments/components-web#step-2-add-components): according to the docs, the full response from `/paymentMethods` should be used in the configuration. With the current implementation as shown above, `paymentMethod.extensions.adyenData.paymentMethodsResponse.paymentMethods` only returns a single PM, moreover it returns an object instead of an array. 
We can refactor `paymentMethod.extensions.adyenData.paymentMethodsResponse` to only return a single object for each PM type but in the end the client side would still have to make an additional `/paymentMethods` call to get the full response and initiate the component.
The extension `paymentMethod.extensions.adyenData.paymentMethodsResponse` can still be useful when looping though each PM to create and mount the payment method.

AndreasA: @peterojo Shouldn't it be enought to provide ALL payment methods for that one payment method in the correct format because anything else is actually really weird. I don't want to be able to switch to Klarna in the web component if I selected credit card and so far it is working fine this way in our implementation.

However, if that is the case, then a separate call the the payment methods response route will be necessary because the payment methods route only returns an array as far as I know.

AndreasA: @peterojo Ah. I might have forgotten the following in my code implementaiton:

```
    $paymentMethodsResponse['paymentMethods'] = array_values(array_filter(
            $paymentMethodsResponse['paymentMethods'],
            function ($value) use ($type) {
                return ($value['type'] ?? null) == $type;
            }
        ));
```

it should be like this I think.

AndreasA: @peterojo If this doesn't work. I guess we need to remove the extension for payment method again and just always call the full payment methods response controller. 

However, personally I think it really does not make sense to retrieve the payment methods response from adyen to pass it on to adyen again with no modifications whatsoever. That is kind of weird :smile:

peterojo: Otherwise we can also remove the extension and let the client side filter each paymentMethod object from the full response (like we are doing in the plugin)

AndreasA: @peterojo Ah. OK. might have misunderstood some parts of the required Adyen web component implementation. However, if I understand your response correctly, the extension should still be used and if it only provides the type but I think it should also provide the payment method config to move as much logic as possible into the Adyen plugin.

- so the full payment methods response as is needed for initialization is provided using the custom route.
- The extension should provide the `details´  for the first payment method and the type which is retrieved from the payment method response be filtering out corresponding data. Before mounting this configuration might be slightly modified to e.g. include the `onChange` handler.

In this case I am all for your change to use `null` or the corresponding object because I would like to remove as much logic as possible from a PWA.

So in your example above:

**Step 1: Initialize adyen checkout**

```
const ADYEN_CHECKOUT_CONFIG = {
    locale,
    clientKey,
    environment,
    showPayButton: false,
    hasHolderName: true,
    paymentMethodsResponse,
    onAdditionalDetails
};
const adyenCheckout = new AdyenCheckout(ADYEN_CHECKOUT_CONFIG);
````

**Step 2: Initialize payment method***
```
adyenCheckout.create(paymentMethod.extensions.adyenData.type, paymentMethod.extensions.adyenData.paymentMethodConfig).mount('#payment-method-container');
```

correct? In wich case you definitely should adjust it to not provide an array but just what is needed.

In this example `paymentMethod.extensions.adyenData.paymentMethodConfig` would be the corresponding payment method config part.


One question which is unrlated to the PR: 
- Can I use the same `AdyenCheckout` object for multiple payment methods or should I create a separate object per payment method. Currently, I create one per payment method but I am not sure this is actually necessary?

peterojo: @AndreasA yes now you understand exactly what I am saying, also I agree with renaming the extension property to `paymentMethodConfig`.
> Can I use the same AdyenCheckout object for multiple payment methods or should I create a separate object per payment method.
Yes you can reuse the same AdyenCheckout object, this is what we do in the plugin as well.

AndreasA: @peterojo Yeah.  Then go ahead and adjust the code and rename the property. It makes way more sense that way. We are not using it that way in our project anyway, as we currently have a working solution and did not update yet. Will probably happen in the near futurre though, so it would be nice if the corresponding change is in it then with an official release tag :smile: 

Regarding reusing AdyenCheckout object: Great, so even if different methods are used, the same adyen checkout object is used. Now it also makes more sense to pass on the full payment methods response object because for multiple payment methods the same response object is used. 

Might be nice, if the adyen web component documentation is made a bit clearer on these things as it might help future developers but it is not a must have.

PR #163, opened 2021-04-19
Add client key, environment and has payment state data to context store api response

AndreasA:
## Summary
Added client key to Store API response, so it can be used by a PWA.

### DIFF
--- a/src/Struct/ContextAdyenDataStruct.php
+++ b/src/Struct/ContextAdyenDataStruct.php
@@ -15,7 +15,7 @@
  *
  * Adyen Payment Module
  *
- * Copyright (c) 2020 Adyen B.V.
+ * Copyright (c) 2021 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  *
--- a/src/Struct/ContextAdyenDataStruct.php
+++ b/src/Struct/ContextAdyenDataStruct.php
@@ -1,0 +1,75 @@
+<?php declare(strict_types=1);
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2020 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\Struct;
+
+use Shopware\Core\Framework\Struct\Struct;
+
+class ContextAdyenDataStruct extends Struct
+{
+    /**
+     * @var string|null
+     */
+    protected $clientKey = null;
+
+    /**
+     * @var string|null
+     */
+    protected $environment = null;
+
+    /**
+     * @var bool
+     */
+    protected $hasPaymentStateData = false;
+
+    public function getClientKey(): ?string
+    {
+        return $this->clientKey;
+    }
+
+    public function setClientKey(?string $clientKey): void
+    {
+        $this->clientKey = $clientKey;
+    }
+
+    public function getEnvironment(): ?string
+    {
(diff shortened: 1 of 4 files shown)


src/Subscriber/ContextResolvedSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+        $salesChannelId = $context->getSalesChannelId();
+
+        $extension = new ContextAdyenDataStruct();
+        $context->addExtension('adyenData', $extension);
+
+        $extension->setClientKey($this->configurationService->getClientKey($salesChannelId) ?: null);
+        $extension->setEnvironment($this->configurationService->getEnvironment($salesChannelId));
+
+        $data = $this->paymentStateDataService->getPaymentStateDataFromContextToken($context->getToken());
+        $extension->setHasPaymentStateData(!empty($data));
```
AndreasA, line 72: @peterojo Just a useful information to have as e.g. if one changes the payment method one has to reload the context and it is good to know if there is adyen data or not.

peterojo: Looks good @AndreasA

src/Subscriber/ContextResolvedSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2020 Adyen B.V.
```
peterojo, line 18:
```suggestion
 * Copyright (c) 2021 Adyen B.V.
```

src/Struct/ContextAdyenDataStruct.php
```diff
@@ -0,0 +1,75 @@
...
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2020 Adyen B.V.
```
peterojo, line 18:
```suggestion
 * Copyright (c) 2021 Adyen B.V.
```

peterojo (approved):

src/Struct/ContextAdyenDataStruct.php
```diff
@@ -0,0 +1,75 @@
...
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <shopware@adyen.com>
+ */
+
+namespace Adyen\Shopware\Struct;
+
+use Shopware\Core\Framework\Struct\Struct;
+
+class ContextAdyenDataStruct extends Struct
```
peterojo, line 29: Lets align the names of these struct extensions, like in #162 , let's call this `AdyenContextDataStruct`

AndreasA: @peterojo done. I thought I used the same name scheme but apparently not :smile:

src/Subscriber/ContextResolvedSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+
+    public function addAdyenData(SalesChannelContextResolvedEvent $event): void
+    {
+        $context = $event->getSalesChannelContext();
+        $salesChannelId = $context->getSalesChannelId();
+
+        $extension = new AdyenContextDataStruct();
+        $context->addExtension('adyenData', $extension);
+
+        $extension->setClientKey($this->configurationService->getClientKey($salesChannelId) ?: null);
```
peterojo, line 68:
```suggestion
        $extension->setClientKey($this->configurationService->getClientKey($salesChannelId));
```
`?:null` is not necessary since `getClientKey` also returns `null`

AndreasA: @peterojo 
not always apparently. I removed the client key from the `All sales channel configuration` and then it returned an empty string. I think it is because I once had a valid value stored there. I could move the `?:` into the getClientKey method though ?

peterojo: @AndreasA no this is fine then :)

peterojo (dismissed):

src/Subscriber/ContextResolvedSubscriber.php
acampos1916: I can confirm this happened to me once too, removing the config values leaves an empty string in the DB apparently

acampos1916:
Thanks for this contribution @AndreasA

acampos1916 (dismissed):

peterojo (approved):

acampos1916 (approved):

PR #164, opened 2021-04-19
[PW-4514] Amazon Pay button

peterojo:
## Summary

Screen Shot 2021-04-22 at 11 16 37 AM
Add support for Amazon Pay button in checkout.
AmazonPay implementation includes a redirect to Amazon after clicking the button in checkout, and then handling the response using the amazon session id on returning to the same checkout page. 
## Tested scenarios

Pay (successfully, failing) with Amazon pay button
Complete failed payment with Amazon pay button
Pay with other pay buttons, google pay and paypal
pay with credit card


src/AdyenPaymentShopware6.php
```diff
@@ -267,6 +265,11 @@ private function updateTo160(UpdateContext $updateContext): void
             );
         }
     }
+
+    private function updateTo200(UpdateContext $updateContext): void
+    {
+        // todo add amazon pay to the next release version
```
acampos1916, line 271: I think we can add this now right? Why would you like to wait?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -16,12 +16,11 @@ export default class ConfirmOrderPlugin extends Plugin {
             locale,
             clientKey,
             environment,
-            showPayButton: false,
+            showPayButton: true,
```
acampos1916, line 19: So this is affecting all PMs now?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -224,8 +225,12 @@ export default class ConfirmOrderPlugin extends Plugin {
             return;
         }
 
+        if (!!componentConfig.prePayRedirect) {
```
acampos1916, line 228: Nice!

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -280,12 +284,55 @@ export default class ConfirmOrderPlugin extends Plugin {
...
+            selectedPaymentMethodObject.type,
+            PRE_PAY_BUTTON
+        );
+        this.mountPaymentButton(paymentMethodInstance);
+    }
+
+    completePendingPayment(paymentMethodType, config) {
+        const url = new URL(location.href);
+        // Check for pending payment session
+        if (url.searchParams.has('pending')) {
```
acampos1916, line 316: Can we use 'pending' as a config constant like config.sessionKey?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: this is only for the AdyenCheckout instance in the confirm page itself, it only renders the payment buttons and handles additionalActions, and actually the `showPayButton` value is already overwritten in the payment button config used to create the button, I am just moving it up into the AdyenCheckout object. There is a separate AdyenCheckout object declared in `checkout.plugin.js` which handles the other PMs

src/AdyenPaymentShopware6.php
peterojo: You're right, no reason to wait for the release PR

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: the original idea was to use `pending` as a generic constant across all components, but actually since I'm passing the component config into this method, I think I can get rid of `pending` and simply check for the `config.sessionKey` instead

acampos1916: That's better yes

acampos1916 (approved):

msilvagarcia (approved):

PR #165, opened 2021-04-24
Static code analysis refactoring

peterojo:
## Summary

Minor updates in response to marketplace warnings:
Screen Shot 2021-04-24 at 11 46 51 PM

## Tested scenarios

All available PMs, including CC 3ds 1&2, wallet payments

**Fixed issue**:


src/Controller/StoreApiController.php
```diff
@@ -51,16 +49,17 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
 
-class StoreApiController extends AbstractStoreController
```
peterojo, line 54: `AbstractStoreController` marked as [@internal](https://github.com/shopware/platform/blob/trunk/src/Core/Framework/Store/Api/AbstractStoreController.php#L16) and moreover we do not use its methods and shopware does not require extending any parent API controller

src/Controller/StoreApiController.php
```diff
@@ -51,16 +49,17 @@
...
+class StoreApiController
 {
     /**
      * @var PaymentMethodsService
      */
     private $paymentMethodsService;
-    /**
-     * @var SalesChannelRepository
-     */
-    private $salesChannelRepository;
```
peterojo, line 63: Unused property

src/Service/CheckoutService.php
```diff
@@ -34,25 +33,10 @@ class CheckoutService extends Checkout
...
-
-    public function startClient($salesChannelId)
-    {
-        try {
-            parent::__construct($this->client->getClient($salesChannelId));
-        } catch (AdyenException $e) {
-            $this->logger->error($e->getMessage());
-        }
+        parent::__construct($this->client);
     }
```
peterojo, line 41: there's no need to declare this service in the DI container since we always needed to call `startClient()` with the current context's salesChannelId, we can simplify this by instantiating it injecting the right client at runtime instead.

src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -309,7 +312,6 @@ public function finalize(
         SalesChannelContext $salesChannelContext
     ): void {
         $transactionId = $transaction->getOrderTransaction()->getId();
-        $this->checkoutService->startClient($salesChannelContext->getSalesChannel()->getId());
```
peterojo, line 312: this line was unnecessary, the checkout service is instead used in src/Service/PaymentDetailsService.php which is called from `resultHandler->processResult()`

src/Service/ClientService.php
```diff
@@ -56,11 +56,6 @@ class ClientService extends Client
      */
     private $apiLogger;
 
-    /**
-     * @var ContainerParametersService
-     */
-    private $containerParametersService;
-
```
peterojo, line 63: unused property

src/Handlers/AbstractPaymentMethodHandler.php
acampos1916: I remember this was added as a fix for multiple sales channels configurations, could you confirm if indeed is not needed anymore?

peterojo: we are still taking the sales channel ID at runtime to create the client, the only change I made is to instantiate Checkout with that client instead of calling startClient

acampos1916 (approved):

msilvagarcia (approved):

PR #180, opened 2021-06-03
[PW-4756] Fix checkout flow to support 6.4

peterojo:
## Summary

Remove mounting of web component in checkout PM list
Mount component on submit order in modal, same as 3ds2 modal
Screen Shot 2021-06-03 at 5 45 54 PM
Screen Shot 2021-06-03 at 5 46 51 PM
Screen Shot 2021-06-03 at 5 34 40 PM

## Tested scenarios

Klarna Pay Now
Updatable PMs (iDeal, CC, 3ds1&2, Bancontact)
Stored PM
3DS2
eWallets

**Fixed issue**:


src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -367,10 +402,40 @@ export default class ConfirmOrderPlugin extends Plugin {
...
+                this.paymentComponent.find('[data-adyen-payment-container]').hide();
+                if (state.isValid) {
+                    let extraParams = {
+                        stateData: JSON.stringify(state.data)
+                    };
+                    let formData = FormSerializeUtil.serialize(this.confirmOrderForm);
+                    this.confirmOrder(formData, extraParams);
+                } else {
+                    component.showValidation();
+                    console.log('Payment failed: ', state);
```
acampos1916, line 425: I like this but maybe we can enable it only when on test mode?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -367,10 +402,40 @@ export default class ConfirmOrderPlugin extends Plugin {
...
+                }
+            }.bind(this)
+        });
+        if (!isOneClick && paymentMethod.type === 'scheme') {
+            configuration.enableStoreDetails = true;
```
acampos1916, line 430: Should this depend on a config or if the storedPaymentMethod PM is enabled?

src/Resources/views/storefront/component/payment/payment-method.html.twig
```diff
@@ -0,0 +1,85 @@
...
+                                                    </label>
+                                                </div>
+                                                <div class="col-sm-12" data-adyen-stored-payment-method-id="{{ storedPaymentMethod.id }}"></div>
+                                            </div>
+                                            {% if loop.last == false  %}<hr>{% endif%}
+                                        {% endfor %}
+                                    {% endif %}
+                                </div>
+                            </div>
+                        {% endif %}#}
```
acampos1916, line 78: Can be removed?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: Good point, we could add a config variable for it

src/Resources/views/storefront/component/payment/payment-method.html.twig
peterojo: Oops :)

acampos1916 (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
cyattilakiss: Also it can be a setting not just for scheme but also for bcmc method

src/Resources/views/storefront/component/payment/payment-action-modal.html.twig
```diff
@@ -8,7 +8,7 @@
         <div class="modal-content">
             <div class="modal-body">
                 <div data-adyen-payment-action-container>
-                    <div class="adyen-payment-action-container loader" role="status">
+                    <div class="adyen-payment-container loader" role="status">
```
cyattilakiss, line 11: what happens if the 3DS2 modal is closed and we want to retry payment?
I assume in this case if the modal is the same as the modal including the input fields before the action then we won't be able to do so and the page needs to be refreshed manually.

Could you check this scenario if it handles it well?

cyattilakiss (approved):
Nice big PR, good job putting it together :) (I had some questions but overall LGTM)

src/Resources/views/storefront/component/payment/payment-action-modal.html.twig
peterojo: @cyattilakiss there is no 'close' button for the 3ds2 modal and it also does not close when you click outside the modal, the only way to close the modal is to refresh the page, since the order has already been created and the cart has been cleared, this will lead to losing the payment session altogether. Payment will remain as 'in progress' and cannot be completed, until OFFER_CLOSED notification sets the payment as 'failed'. 
Also, there is a different modal `payment-component-modal` with the input fields for the payment data

PR #183, opened 2021-06-16
[PW-4825] implement giftcards

### LINKED ISSUE #187: Implement Gift Card component
Implement gift card component with order API

peterojo:
## Summary

Implement giftcards as standalone payment methods
update PM rendering in checkout to support multiple giftcards (giftcards all have the same `paymentMethod.type`)
update PM filtering to handle giftcards

## Tested scenarios

**Fixed issue**: Resolves #187


src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -109,17 +112,27 @@ export default class ConfirmOrderPlugin extends Plugin {
...
         // Get the payment method object from paymentMethodsResponse
-        let filtered = $.grep(this.adyenCheckout.paymentMethodsResponse.paymentMethods, function(paymentMethod) {
-            return paymentMethod.type === type;
+        let paymentMethodConfigs = $.grep(this.adyenCheckout.paymentMethodsResponse.paymentMethods, function(paymentMethod) {
+            return paymentMethod[identifier] === type;
         });
-        if (filtered.length === 0) {
+        if (paymentMethodConfigs.length === 0) {
+            if (this.adyenCheckout.options.environment === 'test') {
+                console.error('Payment method configuration not found. ', state);
```
acampos1916, line 131: Should we add a visible error message here as well?

acampos1916 (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -109,17 +112,27 @@ export default class ConfirmOrderPlugin extends Plugin {
...
+        }
+
         // Get the payment method object from paymentMethodsResponse
-        let filtered = $.grep(this.adyenCheckout.paymentMethodsResponse.paymentMethods, function(paymentMethod) {
-            return paymentMethod.type === type;
+        let paymentMethodConfigs = $.grep(this.adyenCheckout.paymentMethodsResponse.paymentMethods, function(paymentMethod) {
+            return paymentMethod[identifier] === type;
         });
-        if (filtered.length === 0) {
+        if (paymentMethodConfigs.length === 0) {
```
cyattilakiss, line 129: can the `paymentMethodConfigs.length` be empty?
In case the method is not available in the payment methods response then wouldn't it be already filtered out on the backend?

cyattilakiss (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: We have not yet implemented error messages in the frontend, but yes this is a good idea

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: True, it is not likely to happen. This is just a paranoid check, just in case we mess up and the `this.adyenCheckout` is somehow missing the PM

cyattilakiss: okay, it might be still useful for future debugging purposes then if such a thing happens or if the methods response changes

PR #191, opened 2021-07-15
[PW-4823] prefill shopper details and addresses 

peterojo:
## Summary
Prefill personal details and addresses in components that require this data

## Tested scenarios

Afterpay
Oney (3x, 4x)

**Fixed issue**:


acampos1916 (approved):

Morerice (approved):

src/Resources/app/storefront/src/configuration/adyen.js
```diff
@@ -26,6 +26,7 @@ export default {
         'facilypay_4x', 'facilypay_6x', 'facilypay_10x', 'facilypay_12x', 'afterpay_default', 'ratepay',
         'ratepay_directdebit', 'giftcard'
     ],
+    prefillablePaymentMethods: ['facilypay_3x', 'facilypay_4x', 'facilypay_6x', 'facilypay_10x', 'facilypay_12x', 'afterpay_default'],
```
cyattilakiss, line 29: it is possible to just add it to the main component as well so all the methods get this configuration or is it better to keep it separated per method?

peterojo: I think not all the PMs accept the fields `personalDetails`,`billingAddress` and `deliveryAddress`.  At least going by the fields present in their `details` array in the `/paymentMethods` response. To be honest I don't know what happens if I just set this data for all PMs.

cyattilakiss: This is what we do in other plugins as a recommendation from checkout, they are going to take care of filtering out the not necessary fields in the data config, so this should make our life easier regarding maintenance :)

peterojo: Updated to add the addresses for all components, still works as well as the others

cyattilakiss (approved):

Morerice (approved):

PR #192, opened 2021-07-16
[PW-5037] - Add addressDetails to Amazon Pay

Morerice:
## Summary
Add addressDetails into Amazon Pay https://docs.adyen.com/payment-methods/amazon-pay/web-drop-in

1. Add addressDetails to amazonpay config
2. Copy shipping address into addressDetails

## Tested scenarios
* Successful AmazonPay payment
* Cancelled AmazonPay payment
* Successful card payment


cyattilakiss (dismissed):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -361,6 +361,9 @@ export default class ConfirmOrderPlugin extends Plugin {
     }
 
     renderPrePaymentButton(componentConfig, selectedPaymentMethodObject) {
+        if (selectedPaymentMethodObject.type === 'amazonpay') {
+            componentConfig.extra = this.setAddressDetails(componentConfig.extra);
```
acampos1916, line 365: So will this config break the component if it isn't amazonpay? Thinking we could add it to all PrePayButton methods for simplicity if possible. I think it is only used for amazon at the moment anyway

src/Resources/views/storefront/page/checkout/confirm/confirm-payment.html.twig
```diff
@@ -46,6 +48,12 @@
              data-adyen-plugin-id="{{ adyenFrontendData.pluginId }}"
              data-selected-payment-method-handler="{{ adyenFrontendData.selectedPaymentMethodHandler }}"
              data-selected-payment-method-plugin-id="{{ adyenFrontendData.selectedPaymentMethodPluginId }}"
+             data-name="{{ adyenFrontendData.name }}"
+             data-address-line="{{ adyenFrontendData.addressLine }}"
+             data-city="{{ adyenFrontendData.city }}"
+             data-post-code="{{ adyenFrontendData.postCode }}"
+             data-country-code="{{ adyenFrontendData.countryCode }}"
+             data-phone-number="{{ adyenFrontendData.phoneNumber }}"
```
peterojo, line 56: These fields should be removed now right?

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
Morerice: I'm not sure if it will break, but the `setAddressDetails` contains amazonpay specific logic, to change the `productType` to `PayOnly` if there is no phone number.

src/Resources/views/storefront/page/checkout/confirm/confirm-payment.html.twig
Morerice: Yes, my bad. Removed in https://github.com/Adyen/adyen-shopware6/pull/192/commits/b6b6e6086af01dd9f4b1b24c88fb1e39c08ae3f6

peterojo (approved):

cyattilakiss (approved):

PR #202, opened 2021-08-09
[PW-5135] show safe error messages to shopper

peterojo:
## Summary

Display validation error messages to shopper when they are safe
Screen Shot 2021-08-09 at 12 42 06 PM

Safe error messages displayed include only those that the shopper can change and are not already handled by frontend validation.

List includes: 
- [124 - Invalid phone number](https://docs.adyen.com/development-resources/error-codes#124-invalid-phonenumber) 

## Tested scenarios

Afterpay 
Oney


src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -62,13 +62,15 @@
...
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 abstract class AbstractPaymentMethodHandler
 {
 
     const PROMOTION = 'promotion';
+    const SAFE_ERROR_CODES = ['124'];
```
acampos1916, line 73: Could we add a comment here pointing to the docs where the error codes are documented? So that it can be easily maintained

acampos1916 (commented):
Should we create a separate PR to update the composer.lock?

peterojo:
> Should we create a separate PR to update the composer.lock?

I think I can just wait for the other PRs to be merged and fix it in this PR

src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -695,4 +709,14 @@ private function getProduct(string $productId, Context $context): ProductEntity
...
         return $product;
     }
+
+    private function displaySafeErrorMessages(AdyenException $exception)
+    {
+        if (
+            $exception->getErrorType() == 'validation'
+            && in_array($exception->getAdyenErrorCode(), self::SAFE_ERROR_CODES)
+        ) {
+            $this->session->getFlashBag()->add('warning', $exception->getMessage());
```
cyattilakiss, line 719: do you know maybe if the getMessage always human readable?

Morerice:
> Should we create a separate PR to update the composer.lock?

It should be automatically fixed when https://github.com/Adyen/adyen-shopware6/pull/203#pullrequestreview-726103537 is merged

src/Handlers/AbstractPaymentMethodHandler.php
peterojo: I think the [error messages](https://docs.adyen.com/development-resources/error-codes) from the API are always human readable

src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -695,4 +709,14 @@ private function getProduct(string $productId, Context $context): ProductEntity
 
         return $product;
     }
+
+    private function displaySafeErrorMessages(AdyenException $exception)
+    {
+        if (
+            $exception->getErrorType() == 'validation'
```
Morerice, line 716: Should this be a strict comparison?

Morerice (approved):

acampos1916 (approved):

PR #219, opened 2021-12-08
PW-5203 partial refunds

RokPopov:
## Summary
- update the frontend component to allow the admin to input a value, validate the value
- update the db table to save the total amount that has been refunded
- update the AdminController that is being called by Ajax request to accept the amount as a parameter
- update the refund() method of the RefundService to accept an amount and send it to Adyen

## Tested scenarios
- sent a partial refund which was successful (async and sync flow)
- sent a refund call with a empty value which threw an error as expected
- sent a refund call with a value that was higher than the allowed value which also failed with an error as expected
- async flow in all the cases working correctly

### DIFF
--- a/src/Service/RefundService.php
+++ b/src/Service/RefundService.php
@@ -116,7 +116,7 @@ class RefundService
      * @return array
      * @throws AdyenException
      */
-    public function refund(OrderEntity $order): array
+    public function refund(OrderEntity $order, $inputAmount): array
     {
         $orderTransaction = $this->getAdyenOrderTransactionForRefund($order, self::REFUNDABLE_STATES);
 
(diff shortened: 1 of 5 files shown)


Morerice (dismissed):

src/Service/RefundService.php
```diff
@@ -206,11 +205,12 @@ public function insertAdyenRefund(
         OrderEntity $order,
         string $pspReference,
         string $source,
-        string $status
+        string $status,
+        int $refundAmount
```
peterojo, line 209: Please add this new `refundAmount` parameter to the code above (line 186) which calls this method

peterojo (approved):

Morerice (dismissed):

Morerice (approved):

peterojo (approved):

Issue #231, opened 2022-02-14
Call to undefined method getShopwareVersion() StoreService


erwinn:
**Describe the bug**
Since new Shopware version, we receive this error in the checkout:
`Attempted to call an undefined method named "getShopwareVersion" of class "ContainerKsnNhda\StoreService_93b91c4`
`src/Service/ClientService.php:113`

**Versions**
Shopware version: 6.4.8.1
Plugin version: 3.3.0

**To Reproduce**
1. Put items in basket
2. Go to checkout/confirm page in SW
3. You'll receive an 500 error, see error log

**Expected behavior**
Would expect no error

**Screenshots**
![image](https://user-images.githubusercontent.com/10939597/153913702-c3d3f546-537c-4a6d-b074-198ba5de4c46.png)

Morerice:
Hi @erwinn,

Thanks for opening this issue. It has been fixed on v3.3.1.

Best regards,
Jean
Adyen

PR #232, opened 2022-02-17
changed StoreService to InstanceService to fix getShopwareVersion error

Necriso:
## Summary
getShopwareVersion method is not located in StoreService anymore.
To get this plugin to work again changed StoreService class to InstanceService Class

**Fixed issue**:  #231

### DIFF
--- a/src/Service/ClientService.php
+++ b/src/Service/ClientService.php
@@ -32,7 +32,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\EntityRepositoryInterface;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
 use Shopware\Core\Framework\Plugin\PluginEntity;
-use Shopware\Core\Framework\Store\Services\StoreService;
+use Shopware\Core\Framework\Store\Services\InstanceService;
 use Psr\Cache\CacheItemPoolInterface;
 
 class ClientService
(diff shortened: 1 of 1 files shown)


peterojo:
Hi @Necriso thanks for the contribution, we will test this and review it ASAP

Morerice:
Hi @Necriso,

Thanks for opening this PR. Unfortunately this solution would break compatibility with some Shopware 6 versions, since the InstanceService was only introduced in [v6.4.4.0](https://github.com/shopware/platform/blob/v6.4.4.0/src/Core/Framework/Store/Services/InstanceService.php).

Instead of injecting this service, could you inject the version directly as an argument of the service? (Something like `<argument>%kernel.shopware_version%</argument>` should do the trick)

Thanks,
Jean
Adyen

wwnorden:
Hello @Morerice,

thanks for your reply. We can fix it on our own but first its not safe for further updates and second does it mean, that the current version of adyen plugin is not supporting the latest Shopware release?! Sounds a bit weird but we are sure you will fix it asap or don`t you?

Thanks in advanced

Morerice:
Hi @wwnorden 

Yes we will fix it asap. However, since @Necriso opened the PR initially we wanted to give him a day or two to see if he wanted to change this PR and implement the fix we suggested. If no response will be given we will update it ourselves.

Alternatively, if you agree with my previous suggestion, you could also open a PR to be reviewed and released by us, with the proposed solution.

Best regards,
Jean
Adyen

Morerice:
> Hi @Necriso,
> 
> Thanks for opening this PR. Unfortunately this solution would break compatibility with some Shopware 6 versions, since the InstanceService was only introduced in [v6.4.4.0](https://github.com/shopware/platform/blob/v6.4.4.0/src/Core/Framework/Store/Services/InstanceService.php).
> 
> Instead of injecting this service, could you inject the version directly as an argument of the service? (Something like `<argument>%kernel.shopware_version%</argument>` should do the trick)
> 
> Thanks, Jean Adyen

PR #234 with this solution has been merged and released.

PR #233, opened 2022-02-22
Added CaptureService to delay capture for payments

anothaNj:
## Summary

Adjusted the notification process handler to do a automated delayed capture for klarna. It is checked if the order delivery is in a configured state, if not the notification gets rescheduled after a configured amount of time.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/Resources/config/config.xml
+++ b/src/Resources/config/config.xml
@@ -117,4 +117,24 @@
             </helpText>
         </input-field>
     </card>
+    <card>
+        <title>Delayed Capture</title>
+        <input-field type="bool">
+            <name>active</name>
+            <label>Active</label>
+        </input-field>
+        <input-field type="int">
+            <name>rescheduleTime</name>
+            <label>Reschedule Notification Time</label>
+            <helpText>
+                This is the time to wait before trying the next capture in seconds
+            </helpText>
+            <defaultValue>43200</defaultValue>
+        </input-field>
+        <component name="sw-entity-single-select">
+            <name>orderState</name>
+            <entity>state_machine_state</entity>
+            <label>Order Shipping State</label>
+        </component>
+    </card>
 </config>
--- a/src/Resources/config/config.xml
+++ b/src/Resources/config/config.xml
@@ -131,7 +131,7 @@
             </helpText>
             <defaultValue>43200</defaultValue>
         </input-field>
-        <component name="sw-entity-single-select">
+        <component name="sw-entity-single-select-override">
             <name>orderState</name>
             <entity>state_machine_state</entity>
             <label>Order Shipping State</label>
--- a/src/Resources/config/config.xml
+++ b/src/Resources/config/config.xml
@@ -118,14 +118,14 @@
         </input-field>
     </card>
     <card>
-        <title>Delayed Capture</title>
+        <title>Manual Capture</title>
         <input-field type="bool">
             <name>active</name>
-            <label>Active</label>
+            <label>Enabled</label>
         </input-field>
         <input-field type="int">
             <name>rescheduleTime</name>
-            <label>Reschedule Notification Time</label>
+            <label>Capture Delay</label>
             <helpText>
                 This is the time to wait before trying the next capture in seconds
             </helpText>
--- a/src/Resources/config/config.xml
+++ b/src/Resources/config/config.xml
@@ -120,7 +120,7 @@
     <card>
         <title>Manual Capture</title>
         <input-field type="bool">
-            <name>active</name>
+            <name>manualCaptureEnabled</name>
             <label>Enabled</label>
         </input-field>
         <input-field type="int">
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -42,13 +42,20 @@
         </service>
         <service id="Adyen\Shopware\Service\ContainerParametersService" autowire="true">
         </service>
+        <service id="Adyen\Shopware\Service\CaptureService">
+            <argument type="service" id="Adyen\Shopware\Service\Repository\OrderRepository"/>
+            <argument type="service" id="Adyen\Shopware\Service\ConfigurationService"/>
+            <argument type="service" id="Adyen\Shopware\Service\ClientService"/>
+            <call method="setLogger">
+                <argument key="$logger" type="service" id="monolog.logger.adyen_generic"/>
+            </call>
+        </service>
 
         <!--Commands-->
         <service id="Adyen\Shopware\Command\FetchPaymentMethodLogosCommand">
             <argument type="service" id="Adyen\Shopware\ScheduledTask\FetchPaymentMethodLogosHandler"/>
             <tag name="console.command"/>
         </service>
-
         <!--Event subscribers-->
         <service id="Adyen\Shopware\Subscriber\PaymentSubscriber">
             <argument type="service" id="Adyen\Shopware\Provider\AdyenPluginProvider"/>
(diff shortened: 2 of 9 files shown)


acampos1916:
Hey @anothaNj, thanks for this contribution! We will check it out.

peterojo:
Hi @anothaNj thanks again for the contribution, we have planned to add support for manual capture and this PR already gives us a good start.

peterojo (approved):

src/Resources/config/config.xml
```diff
@@ -117,4 +117,24 @@
...
     </card>
+    <card>
+        <title>Delayed Capture</title>
+        <input-field type="bool">
+            <name>active</name>
+            <label>Active</label>
+        </input-field>
+        <input-field type="int">
+            <name>rescheduleTime</name>
+            <label>Reschedule Notification Time</label>
```
acampos1916, line 128: We might reconsider this configuration field in a follow-up PR, as it doesn't sound too user friendly.

src/ScheduledTask/ProcessNotificationsHandler.php
```diff
@@ -198,6 +205,18 @@ public function run(): void
                 if (!$notification->isSuccess()) {
                     $this->handleFailedNotification($notification, $order, $state);
                 }
+/* gesagt.getan. begin -----------------------------------------------------------------------------------------------*/
```
acampos1916, line 208: Thanks for the indicators, but this can also be seen in the Git history, I suggest removing them to keep the code clean
```suggestion
```

src/ScheduledTask/ProcessNotificationsHandler.php
```diff
@@ -198,6 +205,18 @@ public function run(): void
...
+
+            } catch (CaptureException $e) {
+                $this->logger->warning($e->getMessage(), ['code' => $e->getCode()]);
+
+                $scheduledProcessingTime = $this->captureService->getRescheduleNotificationTime();
+                $this->notificationService->changeNotificationState($notification->getId(), 'processing', false);
+                $this->notificationService->setNotificationSchedule($notification->getId(), $scheduledProcessingTime);
+                $this->logger->debug("Payment notification {$notification->getId()} requeued.");
+                continue;
+/* gesagt.getan. end -------------------------------------------------------------------------------------------------*/
```
acampos1916, line 219: Same as above
```suggestion
```

src/Service/CaptureService.php
```diff
@@ -0,0 +1,155 @@
...
+                ['transactions', 'currency', 'lineItems', 'deliveries', 'deliveries.shippingMethod']
+            );
+
+            if (is_null($order)) {
+                throw new CaptureException('Order with order_number ' . $notification->getMerchantReference() . ' not found.', 1645112772);
+            }
+            try {
+                $client = $this->clientService->getClient($order->getSalesChannelId());
+            } catch (AdyenException|\Exception $e) {
+                throw new CaptureException('Capture not able to retrieve Client.', 1645112503, $e);
```
acampos1916, line 65: For clarity
```suggestion
                throw new CaptureException('Capture service not able to retrieve Client.', 1645112503, $e);
```

src/Service/CaptureService.php
```diff
@@ -0,0 +1,155 @@
...
+        ) {
+            $this->logger->info('Capture for order_number ' . $notification->getMerchantReference() . ' start.');
+            $order = $this->orderRepository->getOrderByOrderNumber(
+                $notification->getMerchantReference(),
+                $context,
+                ['transactions', 'currency', 'lineItems', 'deliveries', 'deliveries.shippingMethod']
+            );
+
+            if (is_null($order)) {
+                throw new CaptureException('Order with order_number ' . $notification->getMerchantReference() . ' not found.', 1645112772);
```
acampos1916, line 60: Are these exception codes random numbers? I'm not familiar with these codes

src/Resources/config/services.xml
```diff
@@ -97,5 +97,15 @@
             <argument type="service"
                       id="Adyen\Shopware\Framework\Cookie\AdyenCookieProvider.inner"/>
         </service>
+
+        <!--gesagt.getan. services-->
```
acampos1916, line 101: I would move this service to "Adyen Services" on line 19 instead

acampos1916 (changes requested):

src/Resources/config/config.xml
```diff
@@ -117,4 +117,24 @@
...
+            <helpText>
+                This is the time to wait before trying the next capture in seconds
+            </helpText>
+            <defaultValue>43200</defaultValue>
+        </input-field>
+        <component name="sw-entity-single-select">
+            <name>orderState</name>
+            <entity>state_machine_state</entity>
```
peterojo, line 136: this entity lists all the states in the state machine (payment, delivery and order states) instead of only the delivery states. So the common options like `open`, `in_progress` and `cancelled` will appear 3 times in the dropdown, all with different state IDs

src/Service/CaptureService.php
anothaNj: Yes timestamps, unique codes to find the exception quicker

src/Resources/config/config.xml
anothaNj: this has been an issue in shopware a longer time apparently, added a quick solution as mentioned [here](https://forum.shopware.com/t/criteria-in-sw-entity-single-select-component-in-der-pluginkonfiguration-setzen/64873/10)

src/Resources/config/services.xml
anothaNj: moved it up

src/Service/CaptureService.php
acampos1916: Got it. What do you think @peterojo?

peterojo: Interesting, but it does not seem very practical to me to have a hardcoded unix timestamp integer attached to each exception case. If we want to introduce error codes I think it would be better to use codes that are a bit more memorable (3 or 4 characters), and then document them in a new `.md` file. For now, I believe we can better search the logs with the exception message string instead. Or do I misunderstand the point of these codes?

src/Resources/config/config.xml
```diff
@@ -117,4 +117,24 @@
             </helpText>
         </input-field>
     </card>
+    <card>
+        <title>Delayed Capture</title>
```
peterojo, line 121: Can we change the wording of these labels to be consistent with our [docs](https://docs.adyen.com/online-payments/capture#manual-capture)?
```suggestion
        <title>Manual Capture</title>
```

src/Resources/config/config.xml
```diff
@@ -117,4 +117,24 @@
...
+        <input-field type="bool">
+            <name>active</name>
+            <label>Active</label>
```
peterojo, line 124:
```suggestion
            <label>Enabled</label>
```

src/Resources/config/config.xml
peterojo, line 128:
```suggestion
            <label>Capture Delay</label>
```

src/Service/CaptureService.php
anothaNj: It's simply to pinpoint the location in the code. Searching for the string works aswell.

peterojo (approved):

acampos1916 (dismissed):

peterojo:
@anothaNj this is almost ready to be merged. Could you have a quick look at the code sniffer errors? Also do you mind removing the exception codes as well? they seem a bit too arbitrary for now and may cause confusion.

src/Resources/config/config.xml
```diff
@@ -117,4 +117,24 @@
             </helpText>
         </input-field>
     </card>
+    <card>
+        <title>Manual Capture</title>
+        <input-field type="bool">
+            <name>active</name>
```
peterojo, line 123: Can we be a bit more specific with this config path?
```suggestion
            <name>manualCaptureEnabled</name>
```

src/Service/ConfigurationService.php
```diff
@@ -195,4 +195,34 @@ public function getClientKey(string $salesChannelId)
...
+
+    /**
+     * @param string|null $salesChannelId
+     * @return array|bool|float|int|string|null
+     */
+    public function getOrderState(string $salesChannelId = null)
+    {
+       return $this->systemConfigService->get(self::BUNDLE_NAME . '.config.orderState', $salesChannelId);
+    }
+    // gesagt.getan. end
```
peterojo, line 227: ... and this as well

src/Service/ConfigurationService.php
```diff
@@ -195,4 +195,34 @@ public function getClientKey(string $salesChannelId)
...
         return $this->getClientKeyTest($salesChannelId);
     }
+
+    // gesagt.getan. begin
+
+    /**
+     * @param string|null $salesChannelId
+     * @return array|bool|float|int|string|null
+     */
+    public function isDelayedCaptureActive(string $salesChannelId = null)
```
peterojo, line 205:
```suggestion
    public function isManualCaptureActive(string $salesChannelId = null)
```

src/Service/CaptureService.php
```diff
@@ -0,0 +1,155 @@
...
+        $this->clientService = $clientService;
+    }
+
+    /**
+     * @throws CaptureException
+     */
+    public function doKlarnaCapture(NotificationEntity $notification, Context $context)
+    {
+        if (
+            $this->configurationService->isDelayedCaptureActive() &&
```
peterojo:
```suggestion
            $this->configurationService->isManualCaptureActive() &&
```

peterojo (approved):
LGTM!

PR #238, opened 2022-03-08
[PW-5193] Add manual capture for open invoice payments

peterojo:
## Summary

- Add creation of manual capture requests from admin section
- Update payment response handler to only set payment to 'authorized' if manual capture applies (only webhook notification sets a payment to 'paid')
- Limit manual capture to open invoice payments for now

#### manual capture enabled
- Open invoice payment
- Non open invoice payment
#### manual capture disabled
- Open invoice payment
- Non open invoice payment


src/Controller/AdminController.php
```diff
@@ -139,6 +156,74 @@ public function check(RequestDataBag $dataBag): JsonResponse
         }
     }
 
+    /**
+     * Send a capture request to the Adyen platform
+     *
+     * @Route(
+     *     "/api/adyen/capture",
```
acampos1916, line 163: Are these routes accessible only to requests coming from the admin?

src/Controller/AdminController.php
```diff
@@ -139,6 +156,74 @@ public function check(RequestDataBag $dataBag): JsonResponse
...
+        } catch (CaptureException $e) {
+            $this->logger->error($e->getMessage());
+
+            return new JsonResponse([
+                'success' => false,
+                'message' => 'adyen.error',
+            ]);
+        }
+
+        return new JsonResponse(['success' => true, 'response' => $results]);
```
acampos1916, line 206: If an exception is thrown, is `$results` still defined?

src/Entity/PaymentCapture/PaymentCaptureEntityDefinition.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return new FieldCollection([
+            (new IdField('id', 'id'))->addFlags(new PrimaryKey(), new Required()),
+            (new FkField(
+                'order_transaction_id',
+                'orderTransactionId',
+                OrderTransactionDefinition::class
+            ))->addFlags(new Required()),
+            new StringField('psp_reference', 'pspReference'),
+            (new IntField('amount', 'amount'))->addFlags(new Required()),
```
acampos1916, line 67: Is the currency required here? Maybe we're getting it from the order_transaction_id relationship

acampos1916 (commented):
Left a few comments.

src/Controller/AdminController.php
peterojo: @acampos1916 yes they are, shopware manages that with the `@RouteScope(scopes={"administration"})` at the top of the controller class

src/Controller/AdminController.php
peterojo: no it is not, but then the exception will be returned and this line will not be executed

src/Entity/PaymentCapture/PaymentCaptureEntityDefinition.php
peterojo: Indeed, we get the currency from the order see [this diff](#diff-15f7f071121bfa3ffa4bf6f5b31db091f12a2ce65acae1642ab42c45e0c9eef7R372)

acampos1916 (approved):

Morerice (approved):

PR #243, opened 2022-03-28
Only show ApplePay in checkout page in Safari browser

RokPopov:
Only show ApplePay component in the checkout page if the browser is Safari, otherwise remove it from the payment methods list.

## Tested scenarios
Tested ApplePay in Safari, where the payment method's component is shown in the checkout page. As expected, in other browsers (Chrome, Mozilla) the component is not shown.


src/Service/PaymentMethodsFilterService.php
```diff
@@ -92,6 +94,11 @@ function ($value) use ($pmHandlerIdentifier) {
                     if (empty($paymentMethodFoundInResponse)) {
                         $originalPaymentMethods->remove($paymentMethodEntity->getId());
                     }
+                    // Remove ApplePay PM if the browser is not Safari
+                } elseif ($pmCode == ApplePayPaymentMethodHandler::getPaymentMethodCode()) {
+                    if ($isSafari !== 1) {
+                        $originalPaymentMethods->remove($paymentMethodEntity->getId());
+                    }
```
peterojo, line 101: I would get rid of the extra `if` block since`$isSafari` is already defined at the top
```suggestion
                } elseif ($pmCode == ApplePayPaymentMethodHandler::getPaymentMethodCode() && $isSafari !== 1) {
                    $originalPaymentMethods->remove($paymentMethodEntity->getId());
```

RokPopov: 🙏

peterojo (approved):

acampos1916 (approved):

PR #258, opened 2022-05-19
[PW-6441] Add PreparedPayment flow for headless integrations

peterojo:
## Summary

1. Create new store-api endpoint for making `/payments` requests. Also create paymentDetails and paymentStatus endpoints for prepared payments
    - Reuse logic currently in `AbstractPaymentMethodHandler`
    - Generate unique reference for payment method request
    - Return pspReference and payment reference to the frontend to be sent in `createOrder` request
2. Update existing payment method handlers (only for those that support manual capture) to additionally implement [PreparedPaymentHandlerInterface](https://github.com/shopware/platform/blob/b47b01cab41f1f6d18262289140a7122ca1e70dc/src/Core/Checkout/Payment/Cart/PaymentHandler/PreparedPaymentHandlerInterface.php) and add the tag `<tag name="shopware.payment.method.prepared"/>` in `payment-methods.xml`

3. Create config option for merchants to choose between prepared/async flow
4. If prepared flow is selected in config, implement `validate` function of the handler
    -  Use `/payment/details` call to make sure payment is Authorized otherwise throw exception.
5.  If prepared flow is selected in config, implement `capture` function of the handler
    - do capture request and update order transaction
    - store payment reference in order transaction customFields
6. If async flow is selected then return early from `validate` and `capture` methods, and if prepared flow is selected then return early from `pay` and `finalize` methods.
6. Update all logic that relies on the merchant reference being the order number to instead fetch order by payment reference.
7. Display payment reference in order detail page and link to the payment in CA
Screen Shot 2022-06-07 at 9 46 27 AM

## Tested scenarios

- Completed order using the payment flow for CC and CC 3DS2

**Fixed issue**:

### DIFF
--- a/src/Handlers/CardsPaymentMethodHandler.php
+++ b/src/Handlers/CardsPaymentMethodHandler.php
@@ -26,8 +26,10 @@
 namespace Adyen\Shopware\Handlers;
 
 use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\AsynchronousPaymentHandlerInterface;
+use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\PreparedPaymentHandlerInterface;
 
-class CardsPaymentMethodHandler extends AbstractPaymentMethodHandler implements AsynchronousPaymentHandlerInterface
+class CardsPaymentMethodHandler extends AbstractPaymentMethodHandler implements
+    AsynchronousPaymentHandlerInterface, PreparedPaymentHandlerInterface
 {
 
     public static function getPaymentMethodCode()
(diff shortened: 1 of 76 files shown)


candemiralp (dismissed):

src/Handlers/CardsPaymentMethodHandler.php
```diff
@@ -26,8 +26,11 @@
 namespace Adyen\Shopware\Handlers;
 
 use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\AsynchronousPaymentHandlerInterface;
+use Shopware\Core\Checkout\Payment\Cart\PaymentHandler\PreparedPaymentHandlerInterface;
 
-class CardsPaymentMethodHandler extends AbstractPaymentMethodHandler implements AsynchronousPaymentHandlerInterface
+class CardsPaymentMethodHandler extends AbstractPaymentMethodHandler implements
+    AsynchronousPaymentHandlerInterface,
+    PreparedPaymentHandlerInterface
```
michaelpaul, line 33: Can we declare this implements only once at `AbstractPaymentMethodHandler` where we actually implement the interface instead of repeating for every handler?

peterojo: good idea, I meant to do this in a previous PR but somehow this time I didn't even think about it 😄

peterojo: Done in 12ac9b3

Morerice (approved):

PR #255, opened 2022-05-06
[PW-6502] - Street name and house number separation with regex

candemiralp:
## Summary

House number and street address separation with regex

## Tested scenarios

Card Payments
Form HPPs (Oney 3x4x)
Redirect HPPs (iDeal, Klarna)

PR #264, opened 2022-06-22
Release 3.5.0

peterojo:
## Feature release
- #271 
- #265 
## Fixes
- #255 
- #254 
- #257 
- #266 
- #272 
- #274 
- #279
- #280 
- #278


src/Resources/app/administration/src/sw-order-detail-override/sw-order-detail.html.twig
```diff
@@ -0,0 +1,7 @@
+{% block sw_order_detail_header_title %}
+    {% parent %}
+{% endblock %}
```
acampos1916, line 3: I believe this isn't needed since it is only calling the parent?

peterojo: True. I'll take it out and do one more test

src/Resources/config/config.xml
```diff
@@ -141,4 +141,15 @@
             <label>Order Shipping State</label>
         </component>
     </card>
+    <card>
+        <title>Payment Flow</title>
+        <input-field type="bool">
```
acampos1916, line 146: Will this be On by default? Would it be enabled for merchants upgrading?

peterojo: No they would have to select it. Default is the regular async flow

peterojo:
We're moving the Prepared payment flow to a release candidate branch for v4, so a revert will be made on the development branch.

candemiralp (approved):

candemiralp (approved):

acampos1916 (approved):

PR #282, opened 2022-08-15
[PW-7083] inline checkout instead popup

sevarozh:
## Summary
- merchant would like to have the UX of Shopware Plugin to be “native” instead of the module popup for credit cards
- disabled checkout button (when cart has 'order block error') must not be replaced via custom pay button (like applepay or googlepay). see also issue in shopware/platform https://github.com/shopware/platform/issues/2649

## Tested scenarios
open checkout/confirm
Adyen Web components must not be rendered in popup but inline under payment method description
Bildschirmfoto 2022-08-15 um 15 12 43


acampos1916 (approved):

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
```diff
@@ -33,14 +33,38 @@ export default class ConfirmOrderPlugin extends Plugin {
...
+                console.error('Adyen payment configuration missing.');
+                return;
+            }
+
+            if (this.selectedAdyenPaymentMethod in adyenConfiguration.componentsWithPayButton) {
+                // replaces confirm button with adyen pay button for paywithgoogle, applepay etc.
+                this.initializeCustomPayButton();
+            }
+
+            if (adyenConfiguration.updatablePaymentMethods.includes(this.selectedAdyenPaymentMethod) && !this.stateData) {
```
peterojo, line 62: I think the `&& !this.stateData` part is not strictly necessary, since the page refreshes when a new payment method is selected, this part will always be `true`

peterojo (approved):
Thanks for the contribution @sevarozh! Looks good :)

src/Resources/app/storefront/src/checkout/confirm-order.plugin.js
peterojo: maybe this could be useful if Shopware ever switches back to NOT refreshing the page after every PM selection. For now though, the whole concept of updatable payment methods is obsolete.

PR #283, opened 2022-08-26
allows to use custom line item types in cart & order

miguelseuster:
## Summary
Allows to use custom line items in cart and order.  In the current case only products and promotions can use in a cart and order. If a custom entity will used, the payment process interrupt with an error.

## Tested scenarios
* create a custom line item with processor like a bonus coupon or voucher
* add a line item to the cart
* run checkout


acampos1916 (approved):
Thanks for this contribution @miguelseuster

peterojo (approved):

PR #298, opened 2022-09-21
[PW-6994] Make sure notification is successfully inserted before returning accepted message to Adyen

RokPopov:
## Summary
Currently, we are not checking whether saving of the notification to the database succeeded or not before returning the `[accepted]` string as an acknowledgement of the webhook back to Adyen. This PR introduces a try-catch block for saving the notification. If it fails, we catch the exception and log an error message to the logs.

This PR also introduces a new `WebhookException` class that can be used for future webhook handling.

## Tested scenarios
- tested a flow where inserting the notification to the db failed, made sure the exception was caught and the error message was logged in `dev.log`
- tested the flow with successful notification insertion to the db


src/Exception/WebhookException.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types=1);
```
peterojo, line 1: What's the reason for this new exception

peterojo (approved):

Morerice (approved):

PR #301, opened 2022-10-07
[PW-7220] skip non product items in capture requests

peterojo:
## Summary

Update capture request parameters to skip promotions and other non-product items similarly to the payment request

## Tested scenarios

- Add promotion to cart and pay with klarna (manual capture enabled)

### DIFF
--- a/src/Service/CaptureService.php
+++ b/src/Service/CaptureService.php
@@ -80,6 +80,8 @@ class CaptureService
         'twint',
     ];
 
+    const REASON_DELIVERY_STATE_MISMATCH = 'DELIVERY_STATE_MISMATCH';
+
     private OrderRepository $orderRepository;
     private OrderTransactionRepository $orderTransactionRepository;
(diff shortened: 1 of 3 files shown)


src/Service/CaptureService.php
```diff
@@ -173,7 +176,9 @@ public function doOpenInvoiceCapture(string $orderNumber, $captureAmount, Contex
                     }
                     $results[] = $response;
                 } else {
-                    throw new CaptureException('Order delivery status does not match configuration');
+                    $exception = new CaptureException('Order delivery status does not match configuration');
+                    $exception->reason = self::REASON_DELIVERY_STATE_MISMATCH;
+                    throw $exception;
```
peterojo, line 181: This scenario is not an error per se, this just means that the merchant needs to move update the delivery state before the capture request can be sent.
If we count this scenario as an error and increment the error count, then the notification will be lost after 3 trials and the payment will never be captured.

candemiralp (approved):

RokPopov (approved):

PR #309, opened 2022-11-04
[PW-7345] Create configuration for selecting frontend domain

RokPopov:
## Summary
This PR creates a new plugin admin configuration field for the merchant to select the preferred domain to be used in `returnUrl` from a list of all the domains in the `sales_channel_domain` table.

## Tested scenarios
- make sure the selected `returnUrl` is added to the `/payments` request and the redirect is successful on that url


src/Service/Repository/SalesChannelRepository.php
```diff
@@ -44,9 +53,13 @@ public function __construct(
     public function getCurrentDomainUrl(SalesChannelContext $context): string
     {
         $criteria = new Criteria();
+        $domainUrlId = $this->configurationService->getDomainUrlId($context->getSalesChannelId());
+        $domainId = $this->configurationService->getDomainUrlId($context->getSalesChannelId()) ?:
```
peterojo, line 57: Can we make this config optional so that existing configurations are not broken?
```suggestion
        $domainId = $this->configurationService->getHreflangDefaultDomainId() ?:
```

src/Resources/config/config.xml
```diff
@@ -200,4 +200,16 @@
             </helpText>
         </component>
     </card>
+    <card>
+        <title>Domain URL</title>
```
peterojo, line 204: Can we add a enabled/disabled toggle for this config? For storefront sales channels we are able to get the active domain from the context, only headless sales channels need this option set

src/Service/Repository/SalesChannelRepository.php
```diff
@@ -44,9 +53,13 @@ public function __construct(
...
+            $context->getDomainId();
```
peterojo, line 58: Please revise

src/Resources/config/config.xml
RokPopov: Thanks for the review @peterojo, this is now added

src/Resources/config/config.xml
```diff
@@ -200,4 +200,20 @@
...
+        <input-field type="bool">
+            <name>enableOverrideDefaultDomain</name>
+            <label>Select default domain URL</label>
+        </input-field>
+        <component name="sw-entity-single-select">
+            <name>returnUrlId</name>
```
candemiralp, line 210: Changing the name to `defaultDomainId` may make it more clear and consistent with the other field.

src/Service/ConfigurationService.php
```diff
@@ -345,4 +346,25 @@ public function getAdyenGivingCharityWebsite(string $salesChannelId)
             $salesChannelId
         );
     }
+
+    /**
+     * @param string $salesChannelId
+     * @return array|mixed|null
+     */
+    public function getDomainUrlId(string $salesChannelId): ?string
```
candemiralp, line 354: According to the previous comment, the name of the method may be changed as `getDefaultDomainId` to make it consistent.

src/Service/ConfigurationService.php
```diff
@@ -345,4 +346,25 @@ public function getAdyenGivingCharityWebsite(string $salesChannelId)
...
+    {
+        return $this->systemConfigService->get(self::BUNDLE_NAME . '.config.returnUrlId', $salesChannelId);
+    }
+
+    /**
+     * @param string $salesChannelId
+     * @return null|bool
+     */
+    public function getEnableOverrideDefaultDomain(string $salesChannelId): ?bool
```
candemiralp, line 363: Since it is checking the `bool` state of the config field `enableOverrideDefaultDomain`, changing the name of the method as `getIsOverrideDefaultDomainEnabled` may be consistent.

PR #324, opened 2022-11-28
[PW-7480] Support cancellation for OFFER_CLOSED and CANCELLED notifications

### LINKED ISSUE #310: [PW-7480] Notification OFFER_CLOSED is not supported by the plugin
**Describe the bug**
Given is the following notification in the adyen_notification table https://shot.alssndro.be/ARd0tK

When running: ./bin/console adyen:process-webhooks -vvv the the logs show the that "OFFER_CLOSED is not supported by the plugin". 

https://shot.alssndro.be/wQ6vOo

This is incorrect. When an OFFER_CLOSED notification is send to the plugin the plugin should update the order payment status to "Cancelled" in Shopware.

**Versions**
Shopware version: 6.4.12.0
Plugin version: 3.7.1

**To Reproduce**
1. Place an order with payment method: Bancontact
2. An order will be created in Shopware
3. Don't complete the 3DS flow (Just close your browser)
4. After a while Adyen will mark the offer as closed. An OFFER_CLOSED notification is send to Shopware
5. The notification is processed but the order payment status is not updated to "Cancelled" in Shopware

**Expected behavior**
When the OFFER_CLOSED notification is processed the order payment status should be updated to "Cancelled" in Shopware

**Additional context**

In src/ScheduledTask/ProcessNotificationsHandler.php line 176 "OFFER_CLOSED" is passed as the notification event code ($notification->getEventCode()). But the src/ScheduledTask/Webhook/WebhookHandlerFactory.php does not recognize this type in the switch on line 80. This causes the switch to fallback to the default: Throwing an exception.

**Solution**

Add a OfferClosedWebhookHandler

Add this to the switch inside src/ScheduledTask/Webhook/WebhookHandlerFactory.php
```
case EventCodes::OFFER_CLOSED:
    $handler = new OfferClosedWebhookHandler();
    break;
```

peterojo:
## Summary

Create logic for cancelling orders when processing OFFER_CLOSED, and CANCELLED notifications.
Webhook module returns `CANCELLED` state for these events when successful and when the order is in cancellable state.
- Add handlers for OFFER_CLOSED and CANCELLED webhook event codes
- Abstract and reuse cancellation logic, also for CancelOrRefundHandler

## Tested scenarios

- Test cancellation of OT after webhook module returns `CANCELLED` state.

**Fixed issue**:  resolves #310


src/ScheduledTask/Webhook/CancelOrRefundWebhookHandler.php
```diff
@@ -35,6 +35,8 @@
 
 class CancelOrRefundWebhookHandler implements WebhookHandlerInterface
 {
+    use CancellableWebhookHandlerTrait;
```
Morerice, line 38: Is there a reason why we decided to use a trait instead of abstracting this functionality to a helper class?

candemiralp (approved):

src/ScheduledTask/Webhook/CancelOrRefundWebhookHandler.php
Morerice: Although I don't really agree that this is cleaner, we can still do this but I think we would need to create another interface something like `CancellableInterface` to be paired with this trait. The interface would require the functions that the trait provides.

This would help easily identify which webhook handlers have the cancellable functionality. What do you think?

peterojo: Just an implementation detail though, but personally I think it's better as a trait than an interface

Morerice: Yeah that's a good point. To go back to my initial point though, for me the biggest issue is that in the trait, the function is basically assuming that there's gonna be an`orderTransactionStateHandler` data member in the class that is using the trait. Whereas if this was in a helper class, this dependency would explicitly be defined in the constructor.

peterojo: Oh yeah that's a good point. I think to fix that I would make the trait's function as dumb as possible and just pass it the handler to use as well. What do you think of that @Morerice ?

peterojo: something like
```php
protected function handleCancelWebhook(
        OrderTransactionEntity $orderTransactionEntity,
        OrderTransactionStateHandler $orderTransactionStateHandler,
        string $state,
        Context $context
    ): void {
        if (PaymentStates::STATE_CANCELLED === $state) {
            $orderTransactionStateHandler->fail($orderTransactionEntity->getId(), $context);
        }
    }
```

Morerice: I think that would be better, but I kinda feel like we're overdoing with abstraction for this case, cause either if we use a helper or a trait, all of these changes will only be to abstract an if condition containing a single function call. Also this if condition will only be useful for the `CancelOrRefund` webhook. What do you think?

peterojo: Resolved offline

Morerice (approved):

PR #328, opened 2022-12-01
[PW-7417] Process partial payment notifications

RokPopov:
## Summary
After the implementation of order API flow for giftcard payments, a shopper can make a partial payment with a native giftcard component and then complete the payment in checkout with another PM.
This PR introduces the processing of the AUTHORISATION and ORDER_CLOSED notifications to manage the order transaction states for partial payments. It is doing so by:
- adding the webhook handler for ORDER_CLOSED events
- creating a new DB table -> `adyen_payment` to a record for every payment (indicated by successful AUTHORISATION notification) when processing webhooks with the mentioned event code. The table consists of the following columns:
  - pspreference
  - original_reference
  - merchant_reference
  - merchant_order_reference
  - order_transaction_id
  - payment_method
  - amount_value
  - amount_currency
  - additional_data
  - capture_mode
  - created_at
  - updated_at
  - foreign key -> order_transaction_id references the id of the order_transaction table
- when processing AUTHORISATION, we create a new record in adyen_payment table and compare webhook amount to the amount of the order transaction
  - if webhook amount covers the order transaction total amount, we update order transaction state (normal, non-partial payment) 
  - if not, we are summing up the amounts of the entries in the adyen_payment table (the ones related to the same order), and then comparing that amount against the order transaction amount. If not the same, continue, if it is, update the state to `paid`

## Tested scenarios
- process partial payment with a giftcard + credit card, make sure the ORDER_CLOSED and correlating AUTHORISATION notifications are processed correctly and the state of the payment is updated accordingly


src/Resources/config/services.xml
```diff
@@ -63,9 +66,11 @@
         </service>
         <service id="Adyen\Shopware\ScheduledTask\Webhook\WebhookHandlerFactory">
             <argument type="service" id="Adyen\Shopware\Service\CaptureService"/>
+            <argument type="service" id="Adyen\Shopware\Service\AdyenPaymentService"/>
             <argument type="service" id="Adyen\Shopware\Service\RefundService"/>
             <argument type="service"
                       id="Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionStateHandler"/>
+            <argument type="service" id="Adyen\Shopware\Service\Repository\AdyenPaymentRepository" />
```
candemiralp, line 73: Calling the repository `AdyenPaymentRepository` here wouldn't be a good idea. `AdyenPaymentService` has already been injected to this class. What do you think about creating the related method in the service file and removing the repository dependency from here?

src/ScheduledTask/ProcessNotificationsHandler.php
```diff
@@ -308,15 +318,25 @@ private function getOrderTransaction($order, $notification, $logContext)
...
 
     /**
-     * @param $notification
+     * @param NotificationEntity $notification
      * @param $context
      * @param $logContext
      * @return \Shopware\Core\Checkout\Order\OrderEntity|null
      */
-    private function getOrder($notification, $context, $logContext)
+    private function getOrder(NotificationEntity $notification, $context, $logContext)
```
candemiralp, line 326: Could you please add the types of the 2nd and 3rd parameters?

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
```diff
@@ -97,10 +106,13 @@ private function handleSuccessfulNotification(
         OrderTransactionEntity $orderTransaction,
         NotificationEntity $notification,
         Context $context
-    ) {
+    ): void {
```
candemiralp, line 109: Could you please add the return and property types to all of the methods in order to be consistent?

src/ScheduledTask/Webhook/OrderClosedWebhookHandler.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * @var OrderTransactionStateHandler
+     */
+    private $orderTransactionStateHandler;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
```
candemiralp, line 56:
```suggestion
```

src/ScheduledTask/Webhook/OrderClosedWebhookHandler.php
```diff
@@ -0,0 +1,113 @@
...
+{
+    /**
+     * @var AdyenPaymentService
+     */
+    private $adyenPaymentService;
+
+    /**
+     * @var AdyenPaymentRepository
+     */
+    private $adyenPaymentRepository;
```
candemiralp, line 45: As follow-up to the first comment related to the repository and the service file, could you please double check this usage?

src/ScheduledTask/Webhook/WebhookHandlerFactory.php
```diff
@@ -107,6 +126,14 @@ public static function create(string $eventCode)
                     self::$refundService
                 );
                 break;
+            case EventCodes::ORDER_CLOSED:
+                $handler = new OrderClosedWebhookHandler(
+                    self::$adyenPaymentService,
+                    self::$adyenPaymentRepository,
```
candemiralp, line 132: Service vs Repository

src/Service/AdyenPaymentService.php
```diff
@@ -0,0 +1,89 @@
...
+
+use Adyen\Shopware\Entity\Notification\NotificationEntity;
+use Adyen\Shopware\Service\Repository\AdyenPaymentRepository;
+use Adyen\Util\Currency;
+use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionEntity;
+use Shopware\Core\Framework\Context;
+
+class AdyenPaymentService
+{
+    protected AdyenPaymentRepository $adyenPaymentRepository;
```
candemiralp, line 35: This type definition is not valid for PHP<v8. Could you please use phpdocs instead?

src/Service/AdyenPaymentService.php
```diff
@@ -0,0 +1,89 @@
...
+            'pspreference' => $notification->getPspreference(),
+            'originalReference' => $notification->getOriginalReference() ?? null,
+            'merchantReference' => $notification->getMerchantReference(),
+            'merchantOrderReference' => json_decode($notification->getAdditionalData())->merchantOrderReference ?? null,
+            'orderTransactionId' => $orderTransaction->getId(),
+            'paymentMethod' => $notification->getPaymentMethod(),
+            'amountValue' => $notification->getAmountValue(),
+            'amountCurrency' => $notification->getAmountCurrency(),
+            'additionalData' => $notification->getAdditionalData(),
+            'captureMode' => $isManualCapture ? 'manual_capture' : 'auto_capture'
```
candemiralp, line 58: How about moving `manual_capture` and `auto_capture` strings to the constants? 
There could be a 3rd option in the future for the manual capture cases  for example `MANUALLY_CAPTURED`.

src/Service/AdyenPaymentService.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $currencyUtil = new Currency();
+        $totalPrice = $orderTransactionEntity->getAmount()->getTotalPrice();
+        $isoCode = $orderTransactionEntity->getOrder()->getCurrency()->getIsoCode();
+        $transactionAmount = $currencyUtil->sanitize($totalPrice, $isoCode);
+
+        if ($amountSum >= $transactionAmount) {
+            return true;
+        }
+        return false;
```
candemiralp, line 87:
```suggestion

        return false;
```

src/ScheduledTask/Webhook/OrderClosedWebhookHandler.php
```diff
@@ -0,0 +1,97 @@
...
+    private function handleSuccessfulNotification(
+        OrderTransactionEntity $orderTransactionEntity,
+        NotificationEntity $notificationEntity,
+        Context $context
+    ): void {
+        if ($this->adyenPaymentService->isFullAmountAuthorized(
+            $notificationEntity->getMerchantReference(),
+            $orderTransactionEntity
+        )) {
+            $this->orderTransactionStateHandler->paid($orderTransactionEntity->getId(), $context);
```
peterojo, line 89: I would add a check for manual capture here as well, and move such transactions to `authorized`

src/Service/NotificationService.php
```diff
@@ -280,6 +280,7 @@ public function calculateScheduledProcessingTime(NotificationEntity $notificatio
                 }
                 break;
             case 'OFFER_CLOSED':
+            case 'ORDER_CLOSED':
```
peterojo, line 283: Why postpone `ORDER_CLOSED` for 30 minutes?

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
```diff
@@ -97,27 +106,39 @@ private function handleSuccessfulNotification(
...
-                ['notification' => $notification->getVars()]
-            );
-            $this->captureService->doOpenInvoiceCapture(
-                $notification->getMerchantReference(),
-                $notification->getAmountValue(),
-                $context
-            );
-        } else {
-            $this->orderTransactionStateHandler->paid($orderTransaction->getId(), $context);
+        if ($transactionAmount == $notification->getAmountValue()) {
```
RokPopov, line 119: @peterojo could you double-check this change of moving the amount comparison before going into processing of the AUTHORISATION notification?

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
```diff
@@ -97,27 +106,39 @@ private function handleSuccessfulNotification(
...
+                    ['notification' => $notification->getVars()]
+                );
+                $this->captureService->doOpenInvoiceCapture(
+                    $notification->getMerchantReference(),
+                    $notification->getAmountValue(),
+                    $context
+                );
+            } else {
+                // check for partial payments
+                if (isset(json_decode($notification->getAdditionalData())->merchantOrderReference)) {
```
RokPopov, line 138: @peterojo here the condition changed as well to see whether we have merchantOrderReference

peterojo (approved):

src/Migration/Migration1669129247AdyenPayment.php
```diff
@@ -0,0 +1,47 @@
...
+        $query = <<<SQL
+            CREATE TABLE IF NOT EXISTS `adyen_payment` (
+                `id` BINARY(16) NOT NULL,
+                `pspreference` VARCHAR(255) DEFAULT NULL COMMENT 'PSP Reference',
+                `original_reference` VARCHAR(255) DEFAULT NULL COMMENT 'Original reference',
+                `merchant_reference` VARCHAR(255) DEFAULT NULL COMMENT 'Merchant reference',
+                `merchant_order_reference` VARCHAR(255) DEFAULT NULL COMMENT 'Merchant order reference',
+                `order_transaction_id` BINARY(16)  not NULL COMMENT 'Order transaction Id',
+                `payment_method` VARCHAR(255) NOT NULL COMMENT 'Payment method',
+                `amount_value` VARCHAR(255) NOT NULL COMMENT 'Amount value',
```
Morerice, line 26: Should the amount value here be stored as a string? Why not an int/float?

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
```diff
@@ -97,27 +106,41 @@ private function handleSuccessfulNotification(
...
-            );
-            $this->captureService->doOpenInvoiceCapture(
-                $notification->getMerchantReference(),
-                $notification->getAmountValue(),
-                $context
-            );
-        } else {
-            $this->orderTransactionStateHandler->paid($orderTransaction->getId(), $context);
+        // check for partial payments
+        if (isset(json_decode($notification->getAdditionalData())->merchantOrderReference)) {
```
Morerice, line 120:
```suggestion
        $merchantOrderReference = json_decode($notification->getAdditionalData())->merchantOrderReference)
        if (isset($merchantOrderReference)) {
```

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
```diff
@@ -97,27 +106,41 @@ private function handleSuccessfulNotification(
...
+            return;
+        }
+
+        if ($transactionAmount == $notification->getAmountValue()) {
```
Morerice, line 124: Are we doing loose comparison (== instead of ===) on purpose here?

src/Service/Repository/AdyenPaymentRepository.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param EntityRepository $repository
+     */
+    public function __construct(
+        EntityRepository $repository
+    ) {
+        $this->repository = $repository;
+    }
+
+    /**
+     * Get all partial payments linked to the order, based on the merchant reference
```
Morerice, line 52: Since the function seems to be returning the merchantReference or null, is this comment still relevant?

Morerice (changes requested):

src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php
RokPopov: We have to keep the loose comparison since we are comparing float to a string

PR #361, opened 2023-03-07
[PW-8262] Fix invalid redirect url

GM-Alex:
## Summary

If the payment response type is has the action type redirect the redirect url of the method `AbstractPaymentMethodHandler::pay` should be the url of the PSP, like Klarna for example, instead of the `payment.adyen.redirect_result` route which should be the target after visiting the PSP page.

## Tested scenarios

Use Klarna for example within a headless context.

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -373,8 +373,14 @@ abstract class AbstractPaymentMethodHandler implements AsynchronousPaymentHandle
             $salesChannelContext
         );
 
+        $action = $response[PaymentResponseHandler::ACTION] ?? [];
+        $redirectUrl = ($action['type'] ?? null) === 'redirect' ?
+            $action['url'] ?? null :
+            null;
+
         // Payment had no error, continue the process
-        return new RedirectResponse($this->getAdyenReturnUrl($transaction, $salesChannelContext));
+        return new RedirectResponse($redirectUrl ?? $this->getAdyenReturnUrl($transaction, $salesChannelContext));
+
     }
 
     /**


GM-Alex:
@acampos1916 @rikterbeek @peterojo @Morerice @AlexandrosMor @candemiralp @RokPopov @raoulritter We would need this for a project with a tight timeline. Could you please have a look at this PR? Since the `AbstractPaymentMethodHandler` is not easily modifiable without nearly copy the whole class it would be nice to have this fix ASAP.

src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -373,8 +373,12 @@ public function pay(
             $salesChannelContext
         );
 
+        $action = $response[PaymentResponseHandler::ACTION] ?? null;
+        $redirectUrl = $action['type'] === 'redirect' ? $action['url'] ?? null : null;
```
candemiralp, line 377: This line could throw `Trying to access array offset on value of type null` if the `$action` is null for PHP versions >=7.4.

candemiralp (changes requested):

candemiralp:
Hello @GM-Alex,

Thank you for raising this issue and creating a PR. First we would like to investigate why it is failing although there is no code change for a while. Then we will try to find an abstract solution. I left a comment on the PR also.

Since this issue is urgent for you, I will escalate the urgency to the product specialist.

Best Regards,
Can

src/Handlers/AbstractPaymentMethodHandler.php
GM-Alex: You are right I changed this.

GM-Alex:
Since this change the plugin works flawlessly with the Shopware PWA and Shopware Frontends, expect the credit cards, but there you provide a good documentation how to implement it. 

Also Shopware does it the same way within their PayPal plugin, see https://github.com/shopware/SwagPayPal/blob/master/src/Checkout/Payment/PayPalPaymentHandler.php#L125 

I guess you are handling the redirect correctly within the twig templates by a js script so this issue didn't popped up earlier.

candemiralp:
Hello @GM-Alex,

Thank you for your suggestion and changing your PR. We created an internal investigation ticket to find out the actual cause of the issue because we haven't got any complaints about this issue before. 

After this investigation we may approve your PR or create a new one depending on the issue/solution.

Best Regards,
Can

GM-Alex:
Hi @peterojo,

I'm afraid to tell you this but you are wrong. In line 382 (https://github.com/Adyen/adyen-shopware6/pull/361/commits/e2b872a37e7d3b85c7549c627f6533189008053a#diff-9c1eb916d49e2b822a34215d39a653eac2a404a5bc2b5a7aaba33e765083b777R382) you can see that we only redirect to the PSP (for example Klarna) if the action type is redirect and the url is set. So payments methods which do not involve a redirect action are directly redirected to `payment.adyen.redirect_result`. In addition payments which require a redirect cannot be finalized without the redirect, so this is obviously false and so your implementation is.

Furthermore I now tested my changes with the storefront and the headless approach and it works every think as expected. In addition the type parameter which defines if a payment requires a redirect and is checked by my changes comes from your module / api endpoints and for example for credit card, a none redirecting payment the type is, surprise, not set to redirect.

So please apologize my open words but it seems that you don't tested my modifications nor have a full understanding of your own plugin. So this behavior is imho very disrespectful to you clients, which can expect a working plugin, and also for contributors. I spend now some hours to prove this, which should be your job...

Please feel free to prove me wrong, by providing a counter example, with a step by step guide where my modification breaks the logic. But I'm nearly 100% sure that you aren't able to do so since I now tested my modifications with the default storefront.

Since this is an urgent modification we need for the upcoming going live of our customer please "really" review this PR again. We also can meet via Zoom, Teams or what ever you preferred to get this fixed asap. Thank you. Again sorry for the harsh words but this treatment feels not well.

GM-Alex:
@peterojo As assumed above you are fixing your faulty implementation via JS, within your ConfirmOrderPlugin. To verify this use my PR and disable the ConfirmOrderPlugin at the main.js file of your plugin, see https://github.com/Adyen/adyen-shopware6/blob/e55dcf65bc0ed1e142e83cec024e27116b74f191/src/Resources/app/storefront/src/main.js#L10 and rebuild the storefront by using `bin/build-storefront.sh`. Do the same without my modifications.

You will see that Klarna works, as expected, with my changes and don't work without it. Also my changes have not effect since you are handle it via your JS, so they also will not break anything. And this also shows that you didn't even tested my modifications...

FYI @acampos1916 @rikterbeek @Morerice @AlexandrosMor @RokPopov @raoulritter @candemiralp

peterojo:
Hi @GM-Alex!
Thanks for your feedback! Apologies for the miscommunication in my earlier response, and I didn't mean any disrespect. Your PR will not break other payment methods in the storefront usage of the plugin, but our documented approach for headless integrations which I have explained earlier is recommended for a reason. We would love to hear your feedback if you encountered any issues when making the additional `payment-status` call to get the Adyen API payment response?
I have reached out through our support agent to schedule a call so we can discuss it further.

GM-Alex:
The issue is that the default implementation from Shopware is not working without it properly.

So what's the reason for the approach for headless integrations which you have explained earlier?

I think not using the way of implementation which is Shopware providing makes no sense here. In addition since this is not breaking anything and works auto of the box with Shopware PWA and Shopware Frontends this is the way to go. Furthermore you would be also able to remove JS code for the storefront implementation with this change. Less code leads to less complexity and less complexity leads to better code. There is imho absolutely no reason to not change it.

PR #367, opened 2023-03-30
[PW-8295] Fix giftcard capture mode for adyen_payment table

candemiralp:
## Summary

Method identifier is being obtained from the webhook's `paymentMethod` object instead of `orderTransaction`. That allows setting different capture modes for partial payments.

## Tested scenarios

- Giftcard payments
- Giftcard + Klarna payments

### DIFF
--- a/src/Service/PluginPaymentMethodsService.php
+++ b/src/Service/PluginPaymentMethodsService.php
@@ -78,7 +78,6 @@ class PluginPaymentMethodsService
         foreach ($pluginPaymentMethods as $pluginPaymentMethod) {
             if ($pluginPaymentMethod->getHandlerIdentifier()::getPaymentMethodCode() === 'giftcard' &&
                 $pluginPaymentMethod->getHandlerIdentifier()::getBrand() === $paymentMethod) {
-
                 return $pluginPaymentMethod->getHandlerIdentifier();
             }
         }
--- a/src/Service/PluginPaymentMethodsService.php
+++ b/src/Service/PluginPaymentMethodsService.php
@@ -69,25 +69,20 @@ class PluginPaymentMethodsService
 
     /**
      * @param string $paymentMethod
-     * @return string
-     * @throws AdyenException
+     * @return string|null
      */
-    public function getHandlerIdentifierFromTxVariant(string $paymentMethod): string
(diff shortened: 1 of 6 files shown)


src/Service/PluginPaymentMethodsService.php
```diff
@@ -0,0 +1,93 @@
...
+        return $paymentMethods->getElements();
+    }
+
+    /**
+     * @param string $paymentMethod
+     * @return string
+     * @throws AdyenException
+     */
+    public function getHandlerIdentifierFromTxVariant(string $paymentMethod): string
+    {
```
peterojo, line 76: this method will not catch edge case variants like `visa_amazonpay`, we might end up going down a rabbit hole trying to map all tx_variants in order to get identify adyen payments.
What do you think @candemiralp ?

src/Service/PluginPaymentMethodsService.php
```diff
@@ -0,0 +1,93 @@
...
+            foreach ($pluginPaymentMethods as $pluginPaymentMethod) {
+                if (($pluginPaymentMethod->getHandlerIdentifier()::getPaymentMethodCode() === 'giftcard' &&
+                    $pluginPaymentMethod->getHandlerIdentifier()::getBrand() === $paymentMethod) ||
+                    $pluginPaymentMethod->getHandlerIdentifier()::getPaymentMethodCode() === $paymentMethod) {
+                    return $pluginPaymentMethod->getHandlerIdentifier();
+                }
+            }
+        }
+
+        throw new AdyenException(sprintf("Payment method %s not found!", $paymentMethod));
```
peterojo, line 91: this exception is not caught in `ProcessNotificationsHandler` so this scenario will break the scheduled task

src/Service/PluginPaymentMethodsService.php
candemiralp: You are totally right, let's discuss it on a call.

src/Service/PluginPaymentMethodsService.php
candemiralp: I tested it and It's been caught by handler's catch block and notification is being marked as done after 5th attempt.

PR #368, opened 2023-04-06
[PW-8294] Fix manual capture flow for giftcards

candemiralp:
## Summary

1. In order to capture a payment, full amount should be authorised first.
2. Capture request can be sent from admin order details if only the payment method supports manual capture, full amount is authorised and required amount is not captured yet.
3. If full amount has not been captured successfully (failed capture webhook), sequential capture requests can be sent.
4. Only `requiredCaptureAmount` can be captured for an order. This `requiredCaptureAmount` is obtained from `AdyenPaymentEntity` with manual capture mode payments only.
5. ORDER_CLOSED webhook was fixed for manual capture to use correct merchant reference and capture amount based on the`requiredCaptureAmount` for that transaction.

Notes
1. Manual capture requires the correct shipment state. If that state isn’t satisfied,
 - Capture attempt from admin order detail page will fail.
 - `AUTHORISATION` notification will be re-scheduled to time according to the `capture_delay` config value. But, order status will be marked as `AUTHORISED`.
 - `ORDER_CLOSED` notification will be re-scheduled to time according to the `capture_delay` config value.
 - `Authorisation` and `ORDER_CLOSED` webhooks are being rescheduled for triggering capture again.
2. Plugin assumes giftcards can’t be captured manually.
3. If capture mode is manual, `AUTHORISATION` notification won’t change the payment state to `PAID` but it will trigger a capture request. Capture webhook will update the payment state.

## Tested scenarios

- Auto capture payments (Cards)
- Manual capture payments (Klarna and Klarna + giftcard)

### DIFF
--- a/src/Service/CaptureService.php
+++ b/src/Service/CaptureService.php
@@ -291,7 +291,7 @@ class CaptureService
 
     private function getLineItemsArray(
         ?OrderLineItemCollection $lineItems,
-                                 $currencyCode
+        string $currencyCode
     ): array {
         $lineItemsArray = [];
         $lineIndex = 0;
@@ -317,7 +317,7 @@ class CaptureService
 
     private function buildCaptureRequest(
         string $originalReference,
-               $captureAmountInMinorUnits,
+        $captureAmountInMinorUnits,
         string $currency,
         string $salesChannelId,
         ?array $additionalData = null
--- a/src/Service/CaptureService.php
+++ b/src/Service/CaptureService.php
@@ -27,6 +27,7 @@ namespace Adyen\Shopware\Service;
 use Adyen\AdyenException;
 use Adyen\Client;
 use Adyen\Service\Modification;
+use Adyen\Shopware\Entity\AdyenPayment\AdyenPaymentEntity;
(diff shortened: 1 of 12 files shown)


src/Service/CaptureService.php
```diff
@@ -339,6 +246,93 @@ public function handleCaptureNotification(
...
+
+    private function getLineItemsArray(
+        ?OrderLineItemCollection $lineItems,
+        string $currencyCode
+    ): array {
+        $lineItemsArray = [];
+        $lineIndex = 0;
+        foreach ($lineItems as $lineItem) {
+            // Skip non-product line items.
+            if (empty($lineItem->getProductId()) || $lineItem->getType() !== LineItem::PRODUCT_LINE_ITEM_TYPE) {
```
peterojo, line 300: We could add support for other line items types, same as in #343

candemiralp: Thank you for the review @peterojo! I pushed the changes.

peterojo (approved):

RokPopov (approved):

PR #369, opened 2023-04-07
[PW-8315] Set shopperInteraction for payments request

candemiralp:
## Summary

Set `shopperInteraction` for payments request.


src/Handlers/AbstractPaymentMethodHandler.php
```diff
@@ -434,7 +438,12 @@ protected function preparePaymentsRequest(
 
         if (!empty($request['storePaymentMethod']) && $request['storePaymentMethod'] === true) {
             $request['recurringProcessingModel'] = 'CardOnFile';
-            $request['shopperInteraction'] = 'Ecommerce';
+        }
+
+        if ($paymentMethodEntity->getHandlerIdentifier() === OneClickPaymentMethodHandler::class) {
```
peterojo, line 443: this method will be called from the payment method handler itself so it should be enough to to the following here
```suggestion
        if (static::getPaymentMethodCode() === 'storedPaymentMethods') {
```

peterojo: or
```suggestion
        if (static::class === OneClickPaymentMethodHandler::class) {
```

candemiralp: Thank you for the review @peterojo.

peterojo (approved):

raoulritter (approved):

PR #374, opened 2023-04-17
[PW-8339] Add unique constraint pspreference to adyen_payment table

candemiralp:
## Summary

Unique constraint added to `adyen_payment` table to block duplicate entries for the same (partial or normal) payment.

`AuthorisationWebhookHandler` updated regarding to this requirement.

## Tested scenarios

- Authorisation webhook processing
- Delayed capture after shipment

### DIFF
--- a/src/Migration/Migration1681716164AlterAdyenPayment.php
+++ b/src/Migration/Migration1681716164AlterAdyenPayment.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace Adyen\Shopware\Migration;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+class Migration1681716164AlterAdyenPayment extends MigrationStep
+{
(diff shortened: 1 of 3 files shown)


src/Migration/Migration1681716164AlterAdyenPayment.php
```diff
@@ -0,0 +1,28 @@
...
+    public function update(Connection $connection): void
+    {
+        $query = <<<SQL
+            ALTER TABLE `adyen_payment` ADD CONSTRAINT `UQ_ADYEN_PAYMENT_PSPREFERENCE` UNIQUE (`pspreference`)
+        SQL;
+
+        $connection->executeUpdate($query);
+    }
+
+    public function updateDestructive(Connection $connection): void
```
RokPopov, line 24: I think this method can be removed from the file?

candemiralp: This method is a part of the auto generated migration class.

peterojo (approved):

PR #431, opened 2023-11-01
[ECP-8719] Fix capture, refund and notifications blocks in admin order detail page

candemiralp:
## Summary

Our plugin extends admin order details page to show refunds, captures, notifications and partial payments of the order. This extended block is not used on Shopware 6.5 anymore. Related twig block is extended with this PR. In addition, old implementation was kept to keep it backward compatible with Shopware 6.5 instead of replacing the twig block.

## Tested scenarios

Following cases were tested on both Shopware 6.5 and 6.4.
- Admin page details  visuals
- Capture and refund modifications
- Notifications data grid display


src/Resources/app/administration/src/component/adyen-payment-capture/adyen-payment-capture.html.twig
```diff
@@ -1,44 +1,91 @@
...
+{#            v-if="captureRequests.length"#}
+{#            :dataSource="captureRequests"#}
+{#            :columns="columns"#}
+{#            :showSelection="false"#}
+{#            :showActions="false"#}
+{#            :isLoading="isLoading">#}
+{#        </sw-data-grid>#}
+{#    </div>#}
+{#    <div v-else>#}
+{#        {{ $tc('adyen.error') }}#}
```
hossam-adyen, line 89: do we need commenting those lines? or we can just delete them?

src/Resources/app/administration/src/helper/versionHelper.js
```diff
@@ -0,0 +1,44 @@
...
+        return a.localeCompare(b, undefined, { numeric: true, sensitivity: "case", caseFirst: "upper" })
+    };
+
+    const version = Shopware.Context.app.config.version;
+
+    if (semverCompare(version, "6.5.0.0") === -1) {
+        return true
+    } else {
+        return false;
+    }
```
hossam-adyen, line 41:
```suggestion
    return semverCompare(version, "6.5.0.0") === -1
```

candemiralp: If the given versions are same, `semverCompare()` returns `0`. Due to this, I can't apply your suggestion because `0` will return false also, but in reality it's version `6.5.0.0`.

src/Resources/app/administration/src/component/adyen-payment-capture/adyen-payment-capture.html.twig
candemiralp: It's a leftover from development. I removed, thanks for the heads-up!

hossam-adyen (approved):

khushboo-singhvi (approved):

PR #433, opened 2023-11-17
[ECP-8568] Configure and enable PHPStan

candemiralp:
## Summary

PHPStan static code analysis has been enabled with custom configuration file.

Analysis can be run via `composer phpstan` command in the plugin source directory.


adyen-integrations-support:
Hello @candemiralp,

Thank you for reaching out to us regarding this.

We reviewed the PHPStan integration introduced in this PR and tried to run the static analysis locally. After a few adjustments, we were able to make it run inside the Shopware test shop.

After these changes, we were able to run PHPStan successfully, and it reported 556 errors, mostly related to:

- Missing return types
- Method calls on potentially null objects
- Mismatched parameter or return types
- Undefined method calls
- Incorrect case in class references
- Invalid iterable types passed to `foreach` or `count()`

Screenshot from 2025-05-23 16-53-51-20250523-145351

If we want to include this in our CI/CD pipeline, additional work would be needed to:

- Refine the PHPStan config for plugin context
- Improve type declarations in the plugin itself
- Refactor parts of the plugin to resolve the errors reported by the tool

Considering the additional work required to successfully integrate the PHPStan tool in our CI/CD pipeline for all plugin versions (V3, V4, and V5), the PR will not be merged. The additional analysis and implementation are required for proper PHPStan introduction in the project, which should be handled via the change request. Right now, we don’t have plans to add new features, but we’ll keep this in the backlog for future consideration.

If you have any questions, feel free to let us know.

Thank you,

Kind regards.

PR #441, opened 2023-12-11
ECP-8327 Allow saving multiple statedata

khushboo-singhvi:
## Summary
Allowing to save multiple statedata for the same token value in DB

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/Controller/StoreApi/OrderApi/OrderApiController.php
+++ b/src/Controller/StoreApi/OrderApi/OrderApiController.php
@@ -186,10 +186,13 @@ class OrderApiController
      * @param Request $request
      * @return JsonResponse
      */
-    public function deleteGiftCardStateData(SalesChannelContext $context, Request $request): JsonResponse
+    public function deleteGiftCardStateData(SalesChannelContext $context, Request $request)
     {
-        $this->paymentStateDataService->deletePaymentStateDataFromContextToken($context->getToken());
-
-        return new JsonResponse(['token' => $context->getToken()]);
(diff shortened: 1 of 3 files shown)


src/Controller/StoreApi/OrderApi/OrderApiController.php
```diff
@@ -186,10 +186,13 @@ public function giftcardStateData(SalesChannelContext $context, Request $request
      * @param Request $request
      * @return JsonResponse
      */
-    public function deleteGiftCardStateData(SalesChannelContext $context, Request $request): JsonResponse
```
peterojo, line 189: We always have to return `JsonResponse` from this method

src/Controller/StoreApi/OrderApi/OrderApiController.php
```diff
@@ -186,10 +186,13 @@ public function giftcardStateData(SalesChannelContext $context, Request $request
...
     {
-        $this->paymentStateDataService->deletePaymentStateDataFromContextToken($context->getToken());
-
-        return new JsonResponse(['token' => $context->getToken()]);
+        $stateData = json_decode($request->request->get('stateData', ''), true);
+        if (!empty($stateData)) {
+            $Id = $stateData['id'];
+            $this->paymentStateDataService->deletePaymentStateData($Id);
+            return new JsonResponse(['token' => $context->getToken()]);
+        }
```
peterojo, line 196: please return an empty `JsonResponse([])` if state data is not available

PR #452, opened 2024-01-19
[ECP-8915] AbstractPaymentMethodHandler NullPointer Exception

### LINKED ISSUE #402: [ECP-8841] AbstractPaymentMethodHandler breaks the payment flow in Shopware checkout
**Describe the bug**

**Versions**
Shopware version: 6.4.20
Plugin version: [3.8.1, ...]

**To Reproduce**
Because our env is behind a restricted environment i will explain in details what to do
1. In shopware admin create a product without a category assigned
2. In the front-end of shopware place that product in cart.
3. Go to checkout
4. Choose delivery time
5. Choose Klarna Pay Later as payment method
6. Place the order
7. It will return with error 'Call to a member function getName() on null'

**Expected behavior**
That order is placed and that you are directed to the Klarna payment platform.

**Screenshots**
image
image

**Additional context**
I found out that it happens in the AbstractPaymentMethodHandler.php file.
In version 3.8.1 and 3.9.0 you have the following code:
```$product->getCategories() ? $product->getCategories()->first()->getName() : null```
$product->getCategories() is always true because it returns a `CategoryCollection`

In the later versions you are doing ```if (!is_null($product->getCategories())) { ```
Which is also true because of the `CategoryCollection` response.

Instead it should be this ```$product->getCategories()->count() > 0```

### LINKED ISSUE #450: [ECP-8915] AbstractPaymentMethodHandler NullPointer Exception for Products without Category/Categories
Error Report for
store.shopware.com/adyenpaymentshopware6:^3.13
file: store.shopware.com/adyenpaymentshopware6/src/Handlers/AbstractPaymentMethodHandler.php:663
Is:
```
if (isset($product) && !is_null($product->getCategories())) {
    $productCategory = $product->getCategories()->first()->getName();
} else {
    $productCategory = null;
}
```
Should:
```
if (isset($product) && !is_null($product->getCategories()) && $product->getCategories()->count() > 0) {
    $productCategory = $product->getCategories()->first()->getName();
} else {
    $productCategory = null;
}
```
Why:
product->getCategories() is a collection but might have no items. Unlike an empty array which is consider null, a collection without items is not null. This leads to an error on the next line, when the first item of an empty collection is taken (in this case null) and throws an php exception when trying to get getName() on null. (

### PR DESCRIPTION
**Pull request by hossam-adyen**
AbstractPaymentMethodHandler NullPointer Exception for Products without Category/Categories #450

## Tested scenarios

Fixes #450 
Fixes #402

### DIFF
--- a/src/Handlers/AbstractPaymentMethodHandler.php
+++ b/src/Handlers/AbstractPaymentMethodHandler.php
@@ -660,7 +660,7 @@ abstract class AbstractPaymentMethodHandler implements AsynchronousPaymentHandle
                     $imageUrl = null;
                 }
 
-                if (isset($product) && !is_null($product->getCategories())) {
+                if (isset($product) && !is_null($product->getCategories()) && $product->getCategories()->count() > 0) {
                     $productCategory = $product->getCategories()->first()->getName();
                 } else {
                     $productCategory = null;

### CODE REVIEW

File: src/Handlers/AbstractPaymentMethodHandler.php [hunk 1]
```diff
@@ -660,7 +660,7 @@ protected function preparePaymentsRequest(
                     $imageUrl = null;
                 }
 
-                if (isset($product) && !is_null($product->getCategories())) {
+                if (isset($product) && !is_null($product->getCategories()) && $product->getCategories()->count() > 0) {
```
candemiralp, line 663:
```suggestion
                if (isset($product) && is_array($product->getCategories()) && !empty($product->getCategories())) {
```

candemiralp (approved):

khushboo-singhvi (approved):

Issue #461, opened 2024-02-16
[ECP-9027]Incorrect Composer Version Constraints: Plugin is Not Compatible With SW Versions Older Than 6.4.15.0


alessandroaussems:
In https://github.com/Adyen/adyen-shopware6/commit/e3aa3f8739724f81af1fc5b11b43780e781601b2 all occurences of EntityRepositoryInterface are replaced with EntityRepository.

But since the Adyen Plugin uses Autowiring this breaks the plugin completely on Shopware versions older than 6.4.15.0. 

EntityRepository is only autowirable from 6.4.15.0 see: https://github.com/shopware/shopware/commit/86dff2dd51303bca72f7b6bc196871226c7f9a33

The composer constraints of the Adyen Plugin are therefor wrong and should be corrected.

khushboo-singhvi:
Hello @alessandroaussems, 

Thank you for raising this concern. We've created an investigation ticket to ensure we have the correct approach to support platform versions older than 6.4.15.0. This will help us in providing better support and compatibility across different versions.

We appreciate your patience and understanding as we work on this issue. If you have any further questions or if there's anything else you'd like to discuss, please feel free to let us know.

Regards
Khushboo

PR #462, opened 2024-02-16
Use Correct Version Constraints

alessandroaussems:
**Fixed issue**:  https://github.com/Adyen/adyen-shopware6/issues/461

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,8 +10,8 @@
     "type": "shopware-platform-plugin",
     "license": "MIT",
     "require": {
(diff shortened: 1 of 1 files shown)


candemiralp:
Hello @alessandroaussems,

Thank you for raising this issue and creating a fix for it at the same time. However, we can't apply this fix, since dropping support of a platform version on a non-major release is against to semantic versioning.

Instead of updating the composer.json file, we've introduced a fix #476 to support any version below Shopware 6.4.15.0 and it will be released on the upcoming release.

Best Regards,
Can

PR #469, opened 2024-03-19
Added PrePaymentDataBuildEvent and implemented dispatch in the abstra…

MattMaes:
…ctPaymentMethodHandler

## Summary
There's a possibility in Shopware to have a custom crediting system, be it a rewards program, customer balance or other things that affect the remaining amount of an order.
When using the adyen plugin at the time of writing, it's impossible to apply this to  the payment as Adyen always sends either the full amount or the partial amount (defined by adyen plugin).

In order to make the above scenario possible, I introduced an event so that a developer can use this hook to make changes accordingly.

## Tested scenarios
- Without changes: orders still work as intended - nothing has changed
- Subscribed to the event and changed the amount: only that amount has to be paid through Adyen

### DIFF
--- a/src/Event/PrePaymentDataBuildEvent.php
+++ b/src/Event/PrePaymentDataBuildEvent.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Adyen\Shopware\Event;
+
+use Shopware\Core\Checkout\Order\OrderEntity;
+use Shopware\Core\Checkout\Payment\Cart\AsyncPaymentTransactionStruct;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+final class PrePaymentDataBuildEvent
+{
+    /**
+     * @var int
+     */
+    private $amount;
+
+    /**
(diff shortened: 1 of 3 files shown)


candemiralp:
Hello @MattMaes,

Thank you for your contribution. Your case seems to be a valid one but I still have some questions about the implementation.

If the plugin only uses the amount coming from `PrePaymentDataBuildEvent` while authorising a payment, this implementation might create an issue on [AuthorisationWebhookHandler](https://github.com/Adyen/adyen-shopware6/blob/ad59378f55d56a28b90361bb5b5eb415da5ea6b9/src/ScheduledTask/Webhook/AuthorisationWebhookHandler.php#L146). Since the transaction amount will not be equal to the notification amount, order status will not be updated.

In this case, I expect transaction amount to be the full amount (Adyen payment + shopper credit) however notification only contains Adyen payment amount.

What would you think about this case from the perspective of webhooks?

Best regards,
Can

sneakyvv:
FYI: this is a new (better if I say so myself) approach => https://github.com/Adyen/adyen-shopware6/pull/516

PR #471, opened 2024-03-29
[ECP-9027] Use autowire on service definitions

### LINKED ISSUE #461: [ECP-9027]Incorrect Composer Version Constraints: Plugin is Not Compatible With SW Versions Older Than 6.4.15.0

candemiralp:
## Summary

This PR introduces the changes related to autowiring of dependency injection definitions. Service definitions have been cleaned-up and unnecessary argument definitions have been removed. Only the services requires explicit definition have arguments under autowiring.

## Tested scenarios

- Installation and payments on Shopware 6.6.0.0


khushboo-singhvi (approved):

src/Resources/config/services/subscribers.xml
```diff
@@ -5,44 +5,25 @@
...
-            <argument type="service" id="Shopware\Core\Checkout\Cart\CartPersister"/>
-            <argument type="service" id="Shopware\Core\Checkout\Cart\CartCalculator"/>
-            <argument type="service" id="Shopware\Core\System\SalesChannel\SalesChannel\ContextSwitchRoute"/>
-            <argument type="service" id="Shopware\Core\System\SalesChannel\Context\SalesChannelContextFactory"/>
-            <argument type="service" id="Adyen\Util\Currency"/>
+        <service id="Adyen\Shopware\Subscriber\PaymentSubscriber" autowire="true">
+            <argument key="$router" type="service" id="Symfony\Component\Routing\RouterInterface"/>
+            <argument key="$requestStack" type="service" id="request_stack"/>
+            <argument key="$cartPersister" type="service" id="Shopware\Core\Checkout\Cart\CartPersister"/>
+            <argument key="$contextSwitchRoute" type="service" id="Shopware\Core\System\SalesChannel\SalesChannel\ContextSwitchRoute"/>
```
peterojo, line 12: Why no autowire these too?

candemiralp: Hi @peterojo, 

Thank you for the review. You can find the explanation below for each class injection.

`ContextSubscriber` implements `EventSubscriberInterface` which expects `AbstractContextSwitchRoute` on the constructor, which can't be auto wirable since it's an abstract class. It's required to inject `ContextSwitchRoute` manually. Similarly, that interface expects `AbstractCartPersister` which requires `CartPersister` to be injected manually.

Therefore, we can't auto wire those dependencies on the given classes.

peterojo (approved):

PR #473, opened 2024-04-04
[ECP-8326] Implement multiple giftcard usage

khushboo-singhvi:
## Summary
Implementing multiple giftcard support for Shopware 6 plugin.
This PR includes 
- #472 
- #444 
- #447 
- #454 
- #459 
- #465

PR #458, opened 2024-02-14
add schedule webhook command

wannevancamp:
The PR add a new command to process schedule webhooks.

In some scenarios Shopware scheduled tasks stops running after restarting php-fpm.
By using console commands via crontab this can be prevented.

Example crontab:
```
* * * * * php bin/console adyen:process-webhooks
* * * * * php bin/console adyen:schedule-webhooks
```

PR #427, opened 2023-10-19
Dummy PR [ECP-8525] Updated model based approach with checkout PHP library models

SushmitaThakur:
## Summary
This version supports model based approach (php library checkout models)  for payment request, response, payment details, gift card payment and donations.

## Tested scenarios
1. Payments for  multiple payment methods
2. Donations
3. GIftcards

**Fixed issue**:

PR #478, opened 2024-04-17
Release v3.15.0

RokPopov:
- #460 
- #476 
- #473 
- #466 
- #455 
- #458 
- #468 
- #427 
- #477 
- #481 
- #483 
- #482 
- #480 
- #484 
- #487 
- #489 
- #490 
- #491


src/Resources/app/storefront/src/scss/base.scss
```diff
@@ -74,3 +75,33 @@
 .adyen-payment-action-container {
     height: 100%;
 }
+
+select{
```
candemiralp, line 79: This CSS selector is too broad and it affects all of the loaded `select` HTML elements. That needs to use id or class.

candemiralp (changes requested):

src/Resources/app/storefront/src/scss/base.scss
candemiralp, line 79:
```suggestion
#giftcardDropdown {
```

khushboo-singhvi (dismissed):

candemiralp (approved):

Issue #479, opened 2024-04-17
Server error when opening a order without a transaction with Adyen payment method


SpiGAndromeda:
**Describe the bug**
When opening an order from the Shopware administration that doesn't use an Adyen payment method, an error is thrown by Shopware:

```
[2024-04-17T11:47:23.390554+00:00] php.CRITICAL: Uncaught Error: Call to a member function getPaymentMethod() on null {"exception":"[object] (Error(code: 0): Call to a member function getPaymentMethod() on null at /shopware/custom/plugins/AdyenPaymentShopware6/src/Controller/AdminController.php:305)"} []
[2024-04-17T11:47:23.390951+00:00] request.CRITICAL: Uncaught PHP Exception Error: "Call to a member function getPaymentMethod() on null" at AdminController.php line 305 {"exception":"[object] (Error(code: 0): Call to a member function getPaymentMethod() on null at /shopware/custom/plugins/AdyenPaymentShopware6/src/Controller/AdminController.php:305)"} []
```
Because this happens async, the admin interface still works seemingly normal.

**Versions**
Shopware version: 6.5.8.7
Plugin version: 3.14.4

**To Reproduce**
Adyen payment methods have be enabled. The error will be thrown and appear in the logs anytime a order is opened that contains no transaction with Adyen payment.

**Additional context**
I already investiagted the problem and found a solution.

This method fetches the first transaction of the order that uses a Adyen payment method:
https://github.com/Adyen/adyen-shopware6/blob/70fefb29fbb26ff79decc4dceee282580d64e344/src/Controller/AdminController.php#L301

It should return false if no matching transaction was found by catching any thrown exception, including calling a method on null. However, calling a method on null doesn't throw an exception but an error. 

The method is called in JS by the adyen-payment-capture component:
https://github.com/Adyen/adyen-shopware6/blob/70fefb29fbb26ff79decc4dceee282580d64e344/src/Resources/app/administration/src/component/adyen-payment-capture/index.js#L111
JS error handling catches the 500 response from Shopware and doesn't render the component ... which is correct in the end.

Easy fix: use `\Throwable` instead of `\Exception` (or `\Error` if only errors should be catched) in the catch clause. However, I sould suggest handling nulls more explicitly because the method expects that there can be nulls and because throwing exceptions is performancy heavy in PHP (https://php.watch/articles/php-exception-performance).

I will create a PR for that, but I wanted to ask if there are other places in the code where this conpect is used.

[...]

PR #480, opened 2024-04-17
Add null check to isManualCaptureEnabled method of AdminController

SpiGAndromeda:
## Summary
The additional null check allows the method to return early of no Adyen transaction was found instead of relying that an exception is thrown. Because the original intend of the try-catch was to catch any exception, the catch clause is changed to catch all Throwables (possible since PHP 7).

I removed the variable from the catch clause in the first place. But because this plugin has to be compatible with Shopware 6.4, it has to be PHP 7.4 compatible. Catch clauses without variable are allowed since PHP 8.

## Tested scenarios
None

**Fixed issue**:  https://github.com/Adyen/adyen-shopware6/issues/479

### DIFF
--- a/src/Controller/AdminController.php
+++ b/src/Controller/AdminController.php
@@ -312,7 +312,7 @@ class AdminController
             return new JsonResponse(
                 $this->captureService->isManualCapture($paymentMethodHandlerIdentifier)
             );
-        } catch (Throwable) {
+        } catch (Throwable $t) {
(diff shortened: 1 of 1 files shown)


candemiralp (approved):

candemiralp (approved):

candemiralp:
Hello @SpiGAndromeda,

Thank you for your contribution. This PR has been merged.

Best regards,
Can

PR #488, opened 2024-05-13
Add full with to Adyen component for mobile

wannevancamp:
**Fixed issue**:  On mobile, the card component is not centered. By adding full width to the components these are centered on mobile as well on desktop.


candemiralp:
Hello @wannevancamp,

Thank you for your contribution to the repository. I tried to reproduce the issue and applied the patch you have provided, but unfortunately I didn't observe any change.

As I see, card component is already centered without the fix.

Screenshot 2024-05-14 at 13 46 10

Could you please provide a screenshot of the issue?

Best regards,
Can

adyen-integrations-support:
Hi @wannevancamp ,

We tried to reproduce the issue and applied the provided fix, but we did not observe any change. The card component behaves the same with or without the suggested CSS update.

Best regards

PR #497, opened 2024-05-17
[ECP-8954] Automate the Shopware Marketplace zip file creation as release asset

SushmitaThakur:
Creates a new Github Action Workflow that gets triggers on new release, creates the zip file of the plugin and uploads it into the latest release as an asset.


.github/workflows/upload-release-asset.yml
```diff
@@ -0,0 +1,37 @@
+on:
+  release:
+    types: [created, edited]
```
candemiralp, line 3: `edited` might be tricky. If `edited` is triggered this workflow again, any text or description change might result in a new file. 

Do we really need to trigger it when a release is updated? Generally, we don't update tag version. That means files will not change in any case on a release.

.github/workflows/upload-release-asset.yml
```diff
@@ -0,0 +1,37 @@
...
+          echo "tag_name=$tag_name" >> $GITHUB_OUTPUT
+          echo "Latest Release ID: $release_id"
+          echo "Latest Release Upload URL: $upload_url"
+          echo "Latest Release Tag name: $tag_name"
+
+      - name:  Upload asset
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ steps.get-release-info.outputs.tag_name }} AdyenPaymentShopware6.zip --clobber
```
candemiralp, line 37:
```suggestion
          gh release upload ${{ steps.get-release-info.outputs.tag_name }} AdyenPaymentShopware6.zip --clobber
          
```

.github/workflows/upload-release-asset.yml
SushmitaThakur: Oh, if that is the case, you are right we don't need it on 'edited'. I assumed that we might have scenarios where release might have file changes.

candemiralp (approved):

khushboo-singhvi (approved):

.github/workflows/upload-release-asset.yml
```diff
@@ -0,0 +1,37 @@
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
```
candemiralp, line 4: `workflow_dispatch` event might create ambiguous behaviours if the latest tag is different than the branch on which dispatch is triggered.
```suggestion
```

candemiralp (approved):

PR #509, opened 2024-06-12
Set modal height to Adyen component height

wannevancamp:
## Summary
When a QR/3DS modal is shown and the Adyen component is bigger than the modal, a scroll bar is shown. When removing the height from the modal, the modal height will automatically adapt to the height of the component.

The height on the modal was introduced when the modal had a header which has since been removed.
See: https://github.com/wannevancamp/adyen-shopware6/commit/88b622df269bce464733e9b408ae238e5e08ca55

## Tested scenarios
Mobile before
Demostore 2024-06-12 09-56-16

Mobile after
Demostore 2024-06-12 09-56-54

Desktop before
Demostore 2024-06-12 09-57-18

Desktop after
Demostore 2024-06-12 09-57-30

### DIFF
--- a/src/Resources/app/storefront/src/scss/modal.scss
+++ b/src/Resources/app/storefront/src/scss/modal.scss
@@ -22,10 +22,6 @@
 
 .adyen-payment-modal .modal-dialog {
   height: calc(100% - 3.5rem);
-
-  .modal-content {
-    height: 80%;
-  }
 }
 
 .adyen-payment-container.loader {


adyen-integrations-support:
Hi,

After applying the provided fix, the QR code modal behaves as presented. 

image

image

However, the fix introduces an unintended side effect on the 3DS authentication modal. A scrollbar appears, even though it was not present before. This affects both desktop and mobile views, as shown in the attached screenshots.

Best regards

PR #514, opened 2024-07-03
Fix invalid method usage for sw version < 6.4.17.0

lx-wnk:
This change resolves an error due to the usage of not available methods for older Shopware versions.

The method `getCurrentContext` was introduced with 6.4.17.0 (see [here](https://github.com/shopware/shopware/commit/c871e77688d679241fbb2ad92c8c3d922fe475f8)).

Before this there was no method like this so I added an early return instead.
This currently breaks a stage system with version 6.4.14.0.

As stated in the [Shopware store](https://store.shopware.com/de/adyen70424242359f/adyen-payments-fuer-shopware-6.html), this plugin is currently compatible with the version range: 6.3.1.1 - 6.5.8.11

### DIFF
--- a/src/Subscriber/ContextSubscriber.php
+++ b/src/Subscriber/ContextSubscriber.php
@@ -73,12 +73,17 @@ class ContextSubscriber implements EventSubscriberInterface
 
     public function onContextRestored(SalesChannelContextRestoredEvent $event): void
     {
+        if (!method_exists($event, 'getCurrentSalesChannelContext')) {
+            return;
+        }
+
         $token = $event->getRestoredSalesChannelContext()->getToken();
         $oldToken = $event->getCurrentSalesChannelContext()->getToken();
 
+
         $stateData = $this->paymentStateDataService->fetchRedeemedGiftCardsFromContextToken($oldToken);
-        foreach ($stateData->getElements() as $statedataArray) {
-            $this->paymentStateDataService->updateStateDataContextToken($statedataArray, $token);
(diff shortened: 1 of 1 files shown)


lx-wnk:
@SushmitaThakur could you look over this PR or redistribute it? This is a necessary fix for versions below 6.4.17.0

khushboo-singhvi:
Hello @lx-wnk,

Thank you for your feedback and for bringing this issue to our attention.

We understand that the usage of the getCurrentContext method, introduced in Shopware version 6.4.17.0, is causing compatibility issues with older versions, specifically with version 6.4.14.0.

At this time, our current focus is on supporting Shopware version 6.6 and above. And this PR is branched out from V4 of the plugin which only supports Shopware6.6+. Therefore, the changes in this PR may not be compatible with older versions as listed.

We encourage you to make the necessary adjustments in your environment to ensure compatibility with the version you are using. We appreciate your understanding and cooperation.

Thank you for your continued support.

Best regards,
Khushboo Singhvi

adyen-integrations-support:
Hi @lx-wnk ,

The issue is already fixed but for the V3 plugin version that is compatible with older versions of Shopware (from 6.4.11.0). For the older Shopware versions (6.4.X and 6.5.X) you should use Adyen plugin releases with version number starting with 3 (Adyen plguin versions 3.X.X and [main-3 branch](https://github.com/Adyen/adyen-shopware6/tree/main-3)).

We will close this ticket without merging.

If you have any further concerns or questions don’t hesitate to contact us.

Best regards

PR #516, opened 2024-07-10
Use transaction amount instead of order amount

sneakyvv:
## Summary

Currently the payment requests are made for the order amount, but this incorrect when using partial payments (which is supported in Shopware, even though it is not fully developed by core functionality). It should be the transaction amount. This is also logical in the sense that each payment is about a transaction, NOT for an order, only indirectly it is.


sneakyvv:
Rebased on develop-3 branch

adyen-integrations-support:
Hi @sneakyvv ,

Changing the Adyen plugin to support partial payments would require a deeper analysis of this feature. At the moment, we don’t have this planned, but we appreciate your suggestion so we might consider it in the future.

If you have any further questions or additional suggestions, please don’t hesitate to contact us.

Best regards

PR #517, opened 2024-07-10
[ECP-9277] Add Support for Billie payment method

SushmitaThakur:
This PR includes payment method class, handlers, dependency injection and other supporting structure for Billie payment method.


src/AdyenPaymentShopware6.php
```diff
@@ -426,16 +426,24 @@ private function updateTo3100(UpdateContext $updateContext): void
 
     private function updateTo3150(UpdateContext $updateContext): void
```
candemiralp, line 427: This payment method needs to be added with a separate function for the target version. For example, it's planned to be released as a part of v3.16.0, you should create a new update method in this class and call this method with a comparison. Version comparison examples can be found in the same class.

src/Handlers/BilliePaymentMethodHandler.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    public static bool $isOpenInvoice = true;
+    public static bool $supportsManualCapture = true;
+    public static bool $supportsPartialCapture = true;
+
+    public static function getPaymentMethodCode()
+    {
+        return 'klarna_b2b';
+    }
+}
```
candemiralp, line 37: It's a best practice to add an single empty line to the end of the class.

src/PaymentMethods/BilliePaymentMethod.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getType(): string
+    {
+        return 'redirect';
+    }
+}
```
candemiralp, line 87: It's a best practice to add an single empty line to the end of the class.

src/AdyenPaymentShopware6.php
```diff
@@ -426,16 +426,24 @@ private function updateTo3100(UpdateContext $updateContext): void
...
+        foreach ([
             new PaymentMethods\GiftCardPaymentMethod(),
-            $updateContext->getContext()
-        );
-        $this->setPaymentMethodIsActive(
-            true,
-            $updateContext->getContext(),
-            new PaymentMethods\GiftcardPaymentMethod()
-        );
+            new PaymentMethods\BilliePaymentMethod()
```
candemiralp, line 435: This payment method needs to be added with a separate function for the target version. For example, it's planned to be released as a part of v3.16.0, you should create a new update method in this class and call this method with a comparison. Version comparison examples can be found in the same class.

candemiralp (changes requested):

khushboo-singhvi (dismissed):

src/AdyenPaymentShopware6.php
khushboo-singhvi, line 1: Hey, won't it be V4.x.x

src/AdyenPaymentShopware6.php
```diff
@@ -461,6 +465,22 @@ private function updateTo3150(UpdateContext $updateContext): void
         }
     }
 
+    private function updateTo3160(UpdateContext $updateContext): void
```
khushboo-singhvi, line 468: Same here?

candemiralp (approved):

PR #537, opened 2024-08-26
Do not cancel partially paid transactions on new payment selection

sneakyvv:
## Summary
When choosing a new payment method, all previous transactions are cancelled (except for the ones that are already cancelled). To support partial payment, for example when using a balance payment system, only payment that are not yet paid, i.e. transactions with a partially paid status, should be cancelled. The already paid amount should also be deducted from the transaction for the newly chosen payment method.

### DIFF
--- a/src/Controller/StoreApi/Payment/PaymentController.php
+++ b/src/Controller/StoreApi/Payment/PaymentController.php
@@ -354,20 +354,36 @@ class PaymentController
         $context->scope(
             Context::SYSTEM_SCOPE,
             function () use ($order, $initialStateId, $orderId, $paymentMethodId, $context): void {
-                if ($order->getTransactions() !== null && $order->getTransactions()->count() >= 1) {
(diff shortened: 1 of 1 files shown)


sneakyvv:
Hi, thanks for the feedback. 
You might be right that I may have missed some states, like for example authorized, but if a "in progress"/"non-final" payment also shouldn't be cancelled, then the current logic of cancelling everything except cancelled is also invalid, isn't it?
As for your question why this feature is needed: it is not for another plugin, but for our own POS system, which allows partial payments, as well as payments via a balance system, which are applied before a payment through Adyen is initiated.

adyen-integrations-support:
Hi @sneakyvv ,

Changing the Adyen plugin to support partial payments would require a deeper analysis of this feature. At the moment, we don’t have this planned, but we appreciate your suggestion so we might consider it in the future. 

If you have any further questions or additional suggestions, please don’t hesitate to contact us.

Best regards

PR #553, opened 2024-09-26
Fixed #552 - Correctly handle tax-free in capture service  and Throwable in notification

### LINKED ISSUE #552: Capture request for invoice payment method always fails for tax-free orders and notifications are stuck in processing due to Throwable not handled
**Describe the bug**
If an order is tax free, the capture request cannot be built for the line items as determining the tax-rate fails with 

`Call to a member function getPercentage() on null` in ProcessNotifications (if done through notifications - same for manual capture).

As this is a `\Throwable` which is not caught by `\Exception`, it will also not be handled by the catch block in the notifications handler
https://github.com/Adyen/adyen-shopware6/blob/develop/src/ScheduledTask/ProcessNotificationsHandler.php#L247 

for v3.15.0 https://github.com/Adyen/adyen-shopware6/blob/3.15.0/src/ScheduledTask/ProcessNotificationsHandler.php#L242

This could be changed to `\Throwable`  as that will catch excpetions an throwables.

Therefore, the notification is stuck in processing state, and the error does not show up in any log. Only if one is running the corresponding command, will it show up in the output.

The error itself occurs due to the line `$lineItem->getPrice()->getTaxRules()->highestRate()->getPercentage() * 10` in the capture service: https://github.com/Adyen/adyen-shopware6/blob/develop/src/Service/CaptureService.php#L341 

for v3.15.0 https://github.com/Adyen/adyen-shopware6/blob/3.15.0/src/Service/CaptureService.php#L341

For tax free orders, the tax rules are empty, therefore, the highest rate is null. In new PHP version, this could be fixed like this:

`($lineItem->getPrice()->getTaxRules()->highestRate()?->getPercentage() ?? 0) * 10`.

So two changes are necessary  (see above) to fully fix the issues.

**Versions**
Shopware version: 6.5.8.14
Plugin version: 3.15.0

AndreasA:
Fixed issue: https://github.com/Adyen/adyen-shopware6/issues/552

### DIFF
--- a/src/ScheduledTask/ProcessNotificationsHandler.php
+++ b/src/ScheduledTask/ProcessNotificationsHandler.php
@@ -244,7 +244,7 @@ class ProcessNotificationsHandler extends ScheduledTaskHandler
                  * It will be marked as done.
                  */
                 $this->logger->info($exception->getMessage(), $logContext);
-            } catch (\Exception $exception) {
+            } catch (\Throwable $exception) {
                 $logContext['errorMessage'] = $exception->getMessage();
                 $this->logger->error('Notification processing failed.', $logContext);
                 $this->logNotificationFailure($notification, $exception->getMessage());
--- a/src/Service/CaptureService.php
+++ b/src/Service/CaptureService.php
@@ -337,11 +337,13 @@ class CaptureService
                 continue;
             }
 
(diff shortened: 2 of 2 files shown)


AndreasA:
closing pr as it should the main issue (though the second part should be fixed as well) has been fixed in 4.2.2

Issue #564, opened 2024-10-10
Webhook - Klarna (Invoice) Authorisation with manual capture partly fails if no originalPspReference exists


AndreasA:
**Describe the bug**
In some rare cases, it is possible that  the payment has been authorised in Klarna, but the redirect to the Shop fails or browser was closed prematurely. 

This results in the payment not setting the original psp reference in that redirect.

Now the authorisation webhook is received, which is fine and the payment state is changed to authorised. However, the check regarding manual capture necessary etc. requires the original psp refernece which is missing.

Therefore, after changing the state to authorised nearly everything else fails and also no capture is possible. The order will not even show the received/processed webhook in the admin. I guess the notification should include some reference? or can this be done without the psp reference somehow?

It should also be possible to capture the payment in such a scenario.

**Versions**
Shopware version: 6.5.8.14
Plugin version: 3.15.0

adyen-integrations-support:
Hi @AndreasA ,

This is just a quick note to inform you that we received your ticket and have already started working on resolving your issue.

If you have any further questions or concerns, please let us know.

Best regards

adyen-integrations-support:
Hi @AndreasA ,

We have successfully reproduced the issue you described. The scenario involved completing a payment with Klarna and forcibly closing the browser tab before the redirection to the shop occurred. As a result, the original PSP reference was missing in both the adyen_payment_response and order_transaction tables. This caused the authorization webhook to fail during the capture process.

We are currently investigating possible solutions to address this scenario. We will keep you updated as we make progress.

Best regards

adyen-integrations-support:
Hi @AndreasA ,

Thank you for sharing this suggestion for improvement and for providing detailed information.

At the moment, this feature is not planned in our current roadmap. However, we will keep your feedback in mind for future updates and enhancements to the plugin.

If you have any further suggestions or questions, feel free to let us know.

Best regards

AndreasA:
Hi @teodoratimoti what is the recommended way to handle such orders?

Because, however unlikely this happens, the customer got an order confirmation as the payment is authorized which suggests to them that everything is fine.
However, for the shop owner it is not possible to capture the payment and therefore the payment cannot be completed.

Not sure if there might be a way to do so in Klarna admin directly but at least in Adyen admin there is no way to capture the payment, checked with someone that encountered that issue.

So the only way would be to either:
- cancel the order and kindly ask the customer to place the order again.
- To re-open the payment and kindle ask the customer to pay again.
[...]

PR #572, opened 2024-10-25
Use pspreference to fetch adyen payment

jasperP98:
## Summary
In the `\Adyen\Shopware\ScheduledTask\Webhook\RefundWebhookHandler::handleSuccessfulNotification` you try to fetch the adyen payment by its original reference but this function `\Adyen\Shopware\Service\AdyenPaymentService::getAdyenPayment` will try to search this via the pspreference.

### DIFF
--- a/src/ScheduledTask/Webhook/RefundWebhookHandler.php
+++ b/src/ScheduledTask/Webhook/RefundWebhookHandler.php
@@ -124,7 +124,7 @@ class RefundWebhookHandler implements WebhookHandlerInterface
 
         try {
             $adyenPayment = $this->adyenPaymentService->getAdyenPayment(
-                $notificationEntity->getOriginalReference()
+                $notificationEntity->getPspreference()
             );
 
             $this->adyenPaymentService->updateTotalRefundedAmount(


AndreasA:
would this fix https://github.com/Adyen/adyen-shopware6/issues/564 ?

sneakyvv:
We're not really sure what the original PSP reference is or when it might differ from the psp reference field, but I guess you guys would know, right?

Either way, this fixes our issue as we're noticing that transactions CAN be found with the psp reference while they CANNOT be found with the original reference, so it seems like an obvious fix, but the deeper WHY is not clear.

adyen-integrations-support:
Hi all,

In the current implementation, the method **handleSuccessfulNotification** retrieves the original Adyen payment entity from the database using the **originalReference** field:

**$adyenPayment = $this->adyenPaymentService->getAdyenPayment(
    $notificationEntity->getOriginalReference()
);**

The proposed change suggests replacing **getOriginalReference()** with **getPspreference()**, which is **incorrect** and could lead to issues.

Difference between **originalReference** and **pspReference**:

**originalReference**:
This value identifies the original payment transaction. All subsequent actions (captures, refunds, etc.) are linked to this original transaction.

**pspReference**:
This is the reference for a **specific** operation, such as a refund, chargeback, or additional capture. Each refund can have a unique **pspReference**, but they all belong to the **same original payment**.

Why **originalReference** must be used:

Refunds are linked to the original payment, not to individual refund requests.

Updating the total refunded amount on the **wrong entity** (a refund PSP reference instead of the original payment) would result in **incorrect refund tracking**.

If **pspReference** is used instead of **originalReference**, the system may:

Fail to find the correct payment entity.

Update an unrelated refund record instead of the main payment record.

In **\Adyen\Shopware\Service\AdyenPaymentService::getAdyenPayment**, the method will find the original payment because the **pspreference** field in that row matches the original payment's reference.

Best regards

PR #663, opened 2025-05-13
Fix merchant reference could be null

wrongspot:
## Summary
if the sandbox account is used on several server instances, it can happen that there are reference ids that do not match the current shop. in that case, $merchantReference is null and then application crashes, because https://github.com/wrongspot/adyen-shopware6/blob/develop/src/Service/Repository/OrderRepository.php#L91 in next line 
`getOrderByOrderNumber` needs as first argument a string

## Tested scenarios

**Fixed issue**:


AleksandarBoljanovic:
This issue has beem addressed in version 5.0.0: https://github.com/Adyen/adyen-shopware6/releases/tag/5.0.0. Therefore, I am closing this PR.

PR #664, opened 2025-05-19
Fix duplicate media file handling

lx-wnk:
## Summary
Fixed an occuring bug where you can't run `adyen:fetch-logos` without errors if an image is unassigned but the media file exists.

This PR fixes this by catching the specific exception and using the existing mediaId


adyen-integrations-support:
Hello @lx-wnk,

Please note that we were unable to reproduce this issue locally with the existing codebase. 
As you can see in the [FetchPaymentMethodLogosHandler](https://github.com/Adyen/adyen-shopware6/blob/develop/src/ScheduledTask/FetchPaymentMethodLogosHandler.php#L124), the plugin already handles this by deleting existing media before creating new entries. We suspect the issue you encountered may be related to insufficient delete permissions on your server.

That said, we've decided to approve these changes as they provide a valuable fallback mechanism in case similar issues occur in the future.

This PR will be closed. The changes will be included in our next release once we've updated all three maintained plugin versions accordingly.

Thank you for your contribution!

Kind regards.

PR #673, opened 2025-05-20
add twig blocks around inline scripts

wrongspot:
## Summary
if you set csp headers, inline scripts are no longer secure enough. 
To make the inline scripts from `<script>` to `<script nonce={{ app.request.attributes.get("_cspNonce") }}>`, you would have to copy 90% of the twig template files to adapt one line of code.

that's why i created the blocks around the inline scripts. so that you can customise exactly these places in the theme without having to duplicate the complete code.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/Resources/views/storefront/component/adyen-giving/adyen-giving-component.html.twig
+++ b/src/Resources/views/storefront/component/adyen-giving/adyen-giving-component.html.twig
@@ -15,9 +15,10 @@
         ></div>
 
         <div id='donation-container'></div>
-
-        <script>
-            var adyenGivingConfiguration = document.querySelector('#adyen-giving-configuration').dataset;
-        </script>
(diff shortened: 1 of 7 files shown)


adyen-integrations-support:
Hello @wrongspot,

Thank you for the PR. This improves the current situation, but the solution is still incomplete.

Providing Twig blocks makes it possible to override the nonce; merchants must still manually add it in their theme. A complete solution would be to include the nonce attribute directly in the plugin templates.

The recommended approach is to add the nonce token to all inline scripts used by the Adyen plugin, so that CSP works out of the box without requiring any template overrides.

This PR will be closed. The changes will be included in our next release once we've updated all three maintained plugin versions accordingly.

Thank you for your contribution and your time.

Kind regards.

PR #742, opened 2025-11-26
Validate all required input fields when accessibility tweaks are active

roomian:
## Summary

When custom required form fields are present in the checkout, e. g. from other plugins, and the ACCESSIBILITY_TWEAKS flag is active, the validity of those fields is not reported to the user properly.

This was fixed in part with v4.3.7, but for the default TOS checkbox only. With this PR, **all** required input fields in the checkout – including the TOS box – will be validated and reported to the user.

field.reportValidity() has to be called in the event listener callback because the validation message wouldn't be shown for the field otherwise. 

## Tested scenarios
**without ACCESSIBILITY_TWEAKS, before this fix:**
- Attempted to finish checkout with TOS box unchecked – scrolls to and focuses TOS box
- Attempted to finish checkout with a required third party form field left empty – scrolls to and focuses field

**with ACCESSIBILITY_TWEAKS, before this fix:**
- Attempted to finish checkout with TOS box unchecked – scrolls to and focuses TOS box 
- Attempted to finish checkout with a required third party form field left empty – validity is not reported to the user, PayPal window opens for a short moment and closes again

**with ACCESSIBILITY_TWEAKS, after this fix:**
- Attempted to finish checkout with TOS box unchecked – scrolls to and focuses TOS box
- Attempted to finish checkout with a required third party form field left empty – scrolls to and focuses field

### DIFF
--- a/src/Resources/app/storefront/src/configuration/adyen.js
+++ b/src/Resources/app/storefront/src/configuration/adyen.js
@@ -19,12 +19,25 @@
  * See the LICENSE file for more info.
  *
  */
-function validateTOS(self) {
-    const tosCheckbox = document.querySelector('#tos');
-    if(adyenCheckoutOptions?.accessibilityTweaks === '1' && tosCheckbox && !tosCheckbox.checked) {
-        tosCheckbox.scrollIntoView({ behavior: 'smooth', block: 'center' });
-        tosCheckbox.focus();
-        return false;
+
+function validateRequired(self) {
+    if (adyenCheckoutOptions?.accessibilityTweaks === "1") {
+        const requiredFields = document.querySelectorAll("input[required]");
+
+        for (const field of requiredFields) {
+            if (!field.checkValidity()) {
+                const onScrollEnd = () => {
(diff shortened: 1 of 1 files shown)


adyen-integrations-support:
Hi @roomian 

This fix was released in plugin version 4.3.8: 

https://github.com/Adyen/adyen-shopware6/releases/tag/4.3.8

validateRequired() now iterates over all input[required] fields in the checkout (including the default TOS checkbox and fields injected by third-party plugins), regardless of whether ACCESSIBILITY_TWEAKS is enabled. On the first invalid field it smooth-scrolls the field into view, focuses it and calls field.reportValidity() so the browser's validation message is actually displayed. Validation returns false, which rejects the wallet component's onClick and prevents the payment sheet from opening with an incomplete form. As a fallback, the whole confirm order form is then checked via confirmOrderForm.checkValidity(), with reportValidity() called on the form if it fails.

We will proceed with closing this ticket.

Thank you for understanding.

Kind regards

Issue #824, opened 2026-05-29
Compatibility with Swag Security 4.0.10


cwittstock:
**Describe the bug**
Since updating the Shopware security plugin, checkout with Adyen no longer works. When clicking "Place order," a 500 server error occurs:

`Uncaught PHP Exception TypeError: "Adyen\Shopware\Service\PaypalPaymentService::__construct(): Argument #6 ($handlePaymentMethodRoute) must be of type Shopware\Core\Checkout\Payment\SalesChannel\HandlePaymentMethodRoute, Swag\Security\Fixes\GHSA9v5m39wh5chq\HandlePaymentMethodRouteDecorator given, called in /var/www/var/cache/prod_h3c389c93d0db8428320dcae59cc04838/ContainerD5ulj1Q/Shopware_Core_KernelProdContainer.php on line 2710" at PaypalPaymentService.php line 71 {"exception":"[object] (TypeError(code: 0): Adyen\\Shopware\\Service\\PaypalPaymentService::__construct(): Argument #6 ($handlePaymentMethodRoute) must be of type Shopware\\Core\\Checkout\\Payment\\SalesChannel\\HandlePaymentMethodRoute, Swag\\Security\\Fixes\\GHSA9v5m39wh5chq\\HandlePaymentMethodRouteDecorator given, called in /var/www/var/cache/prod_h3c389c93d0db8428320dcae59cc04838/ContainerD5ulj1Q/Shopware_Core_KernelProdContainer.php on line 2710 at /var/www/custom/plugins/AdyenPaymentShopware6/src/Service/PaypalPaymentService.php:71)"} []`

Since the advisory (https://github.com/shopware/shopware/security/advisories/GHSA-9v5m-39wh-5chq) is included in the Shopware 6.7.10.1 core, it stands to reason that the Adyen plugin is not compatible with this version.

After unchecking the first checkbox in the settings of the following screenshot, the checkout process works again.
<img width="982" height="433" alt="Image" src="https://github.com/user-attachments/assets/74470bb6-6d06-4892-8335-3e3c4a9a9e9e" />

**Versions**
Shopware version: 6.7.10.0
Plugin version: 5.3.1

sdrwtf:
I have the same issue

- Shopware version 6.5.8.7
- Plugin version 3.21.0

but also plugin version 4.6.1 is affected.

I had a look into it, seems like only the file `src/Service/PaypalPaymentService.php` is causing this. 

### The Problem
[...]

PR #826, opened 2026-05-29
Fix: Use AbstractHandlePaymentMethodRoute to support decorators

sdrwtf:
## Summary
The strict type-hinting of `HandlePaymentMethodRoute` in `PaypalPaymentService` causes a DI container crash (500 TypeError) when a decorator is applied (e.g., by the Shopware Security Plugin for GHSA-9v5m-39wh-5chq). 

Changed the type to `AbstractHandlePaymentMethodRoute` and added Symfony's `#[Autowire]` attribute to ensure the correct default service is injected while maintaining compatibility with decorators. This aligns the file with the approach already used in other controllers like `FrontendProxyController`.

## Tested scenarios
* Tested on a Shopware 6.5.8.7 instance.
* Verified that the DI container compiles successfully **with** the Shopware Security Plugin active (the decorator is accepted).
* Verified that the DI container compiles successfully **without** the Shopware Security Plugin active (the default Shopware service is injected via Autowire).
* Verified that the checkout process completes for Adyen payment methods without throwing 500 errors.

**Fixed issue**:  #824

### DIFF
--- a/src/Service/PaypalPaymentService.php
+++ b/src/Service/PaypalPaymentService.php
@@ -44,11 +44,12 @@ use Shopware\Core\Checkout\Cart\SalesChannel\CartOrderRoute;
 use Shopware\Core\Checkout\Cart\SalesChannel\CartService;
 use Shopware\Core\Checkout\Order\OrderDefinition;
 use Shopware\Core\Checkout\Payment\PaymentException;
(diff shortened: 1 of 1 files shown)


sdrwtf:
Issue #824 was fixed with PR #826 so I close this PR.

